PR #1, opened 2016-05-01
Initial pass at a set of common definitions

geoffjentry:


Definitions.md
```diff
@@ -0,0 +1,16 @@
+#Definition Of Terms
+
+##Job
```
denis-yuen: Is my understanding here that Job is unlike the other three in that unlike Task, Workflow, or Tool, a Job is actually the only term here that is "actually run." In that there could be many Jobs which are effectively invocations of Tasks with different inputs and parameters?

If so, another name for Job (perhaps an alias?) might be TaskInstance which could imply symmetry with WorkflowInstance for example.

pgrosu:
It would be nice to have some examples, especially where definitions could have different interpretations.

Definitions.md
denis-yuen: To provide content for that diagram idea, I'm thinking:
Job/TaskInstance is an instantiation of a Task. 
Workflow is composed of Tasks. 
WorkflowInstance is an instantiation of a Workflow. 
WorkflowInstance is composed of Jobs/TaskInstances. 

Not sure how Tool fits in yet.

denis-yuen:
Examples from our in-house environment, translated to these terms:

A Workflow is a set of binary files, libraries, and dependencies that describe a workflow at rest. This could be zipped up, but expanded it fully describes a set of operations that can be applied to input files that have to be specified by ...
... a WorkflowInstance which is a "live" executing Workflow, kept track of by our software, and has actual parameters and specified inputs/output files. When running, this is broken up into many parts which are actually ...
... jobs/taskInstances which are actual SGE jobs executing on the cluster. When programming though, workflow developers actually interact with a ...
... Task which is the source code representation of a job, written in Java (can also be the source code for a Bash Script).

pgrosu:
To simplify it:

```
       Task ⊂ Workflow
   Workflow ⊂ Tool
  ______________________
       Task ⊂ Tool


     λ{t ∈ Task}.f() ∈ Job
____________________________
 λ{w ∈ Workflow}.f() ∈ Job 
    λ{tl ∈ Tool}.f() ∈ Job 
```

Definitions.md
ntijanic: In case of nested workflows, is the running instance of an inner "nested workflow task" a TaskInstance or WorkflowInstance?

It would be great to have a superclass term for TaskInstance and WorkflowInstance. Can we perhaps use "Job" for this, rather than an alias for TaskInstance?

geoffjentry: @denis-yuen I like that. Regarding `tool` my thought there was that regardless of what we want people use that word to mean "something I run" but the granularity of "something" can be wildly different. My inclusion of it here was more of a nod to it being a fairly generic word that in and of itself doesn't carry any concrete information. Perhaps I could say something to that effect in the block?

geoffjentry: @ntijanic Good point. We could just amend the `Workflow` text to say it can contain workflows or tasks. I don't think we need to worry about concepts like supertypes in a thing like this. In terms of implementation I agree with you though - if the text were changed as I just described and I were going to implement it that's how I'd do it.

geoffjentry:
@denis-yuen I updated the text to morph Job to TaskInstance and fleshed out tool a bit more to make it more clear what I was intending.

geoffjentry:
@pgrosu I love the lambdas :)

pgrosu:
@geoffjentry Yeah they sometimes help me simplify things :)  Maybe the following might be easier to write up than the type theory approach I wrote above, which was more to simplify the language elements.  I am in the middle of playing around with an augmented CWL schema definition for a concept-engine I will write for @mr-c, but a few of those ideas might apply here.   So here is a very rough language definition for the Workflow-Execution-Schemas (which I'll refer to as WES below):

```
(define-type WES
   [Id Symbol]
   [Call WES WES] ; Call WES-Function WES-ARGS
   [Function Symbol TYPE WES TYPE]       ; name, in-type, body, out-type
   [Replace-With Symbol TYPE WES WES] )  ; name, type, renamed-expr, body

(define-type Task
   [Id Symbol]
   [Function Symbol TYPE WES TYPE]) ; name, in-type, body, out-type

(define-type Workflow
   [Id Symbol]
   [Call Task]) ; Call WES-Function WES-ARGS, as a pipeline such as job1 | job2 | job3 is 
                  equivalent to job3(job2(job1))

(define-type Tool
   [Id Symbol]
   [Call Workflow]) ; Call WES-Function WES-ARGS

(define-type Job
   [Id Symbol]
   [Call Task])     ; Call WES-Function WES-ARGS
   [Call Workflow]) ; Call WES-Function WES-ARGS
   [Call Tool])     ; Call WES-Function WES-ARGS

(define-type TYPE
   [BooleanT]
   [NumberT]
   [StringT]        ; [StringT] could just be [ExpressionT]
   [ExpressionT]
   [WesT])
```

The funny thing is that we can simplify things even more as I notice that some of the above seem redundant and some can also be referenced.  If we need a more graph-like workflow, then we can implement `ListOf(...)` and other concepts.

Hope it helps,
~p

PR #4, opened 2016-11-03
WIP: Workflow Execution Schema

nielsdrost:
Hello,

As we needed a way to run CWL using a rest API, we've created a scheme based on the [cwltool-service](https://github.com/common-workflow-language/cwltool-service). As @tetron suggested this could be usefull for the ga4gh community, I made this into a pull request here.

Note that this scheme is close but not identical to the cwltool-service. Biggest difference is the absence of the "pause" state, for which we could not think up an actual use case, nor a sane implementation. We would be happy to hear your thoughts on this.

To test this scheme we also forked [cwltool-service](https://github.com/NLeSC/cwltool-service), and altered the implementation to use connexion.

All work in progress, but I wanted to let you know we are working on this and get your feedback.

### DIFF
--- a/src/main/resources/swagger/ga4gh-workflow-execution.yaml
+++ b/src/main/resources/swagger/ga4gh-workflow-execution.yaml
@@ -1,7 +1,7 @@
 swagger: '2.0'
 info:
   version: 1.0.0
-  title: Compute Service API
+  title: CWL workflow execution service API
   license:
     name: Apache 2.0
     url: http://www.apache.org/licenses/LICENSE-2.0.html
@@ -135,7 +135,7 @@ paths:
           description: Job ID
           required: true
       responses:
-        '200':
+        '204':
           description: Job deleted
         '404':
           description: Job not found
--- a/src/main/resources/swagger/ga4gh-workflow-execution.yaml
+++ b/src/main/resources/swagger/ga4gh-workflow-execution.yaml
@@ -1,7 +1,7 @@
 swagger: '2.0'
 info:
   version: 1.0.0
-  title: CWL workflow execution service API
+  title: GA4GH workflow execution service API
   license:
     name: Apache 2.0
     url: http://www.apache.org/licenses/LICENSE-2.0.html
@@ -18,7 +18,7 @@ paths:
       parameters:
         - in: body
           name: body
-          description: Input binding for cwl workflow.
+          description: Input binding for workflow.
           required: true
           schema:
             $ref: '#/definitions/job-description'
@@ -217,7 +217,7 @@ paths:
         '404':
           description: Job not found
 definitions:
-  cwl-binding:
+  workflow-binding:
     type: object
     additionalProperties: true
   job-description:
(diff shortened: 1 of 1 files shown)


pvanheus:
This is starting to look a bit like a REST version of the DRMAA API (see https://jira.tecgraf.puc-rio.br/confluence/display/CSBase/OpenDreams+REST).

Specifically on the question of pause - since this schema has a notion of a queue of waiting jobs, the "pause" state is analogous to putting a job on hold so that other jobs can skip ahead of it.

src/main/resources/swagger/ga4gh-workflow-execution.yaml
```diff
@@ -0,0 +1,289 @@
+swagger: '2.0'
+info:
+  version: 1.0.0
+  title: CWL workflow execution service API
```
geoffjentry: CWL != GA4GH

geoffjentry:
@nielsdrost Can you explain more what this is intended to be? This appears to be a CWL schema, is it a CWL binding to the workflow execution schema?

nielsdrost:
Thanks all for the comments :-)

This is intended to be a generic REST API for running workflows, where some resource (cluster, cloud, big server) is exposing and endpoint that users (or more likely some web application) can submit (batch) jobs to.

It assumes all the "what will we run" is handled by the workflow specification language used (could be CWL, could be WDL, could be something else entirely). What is left is a simple interface to submit a job, get a list of all jobs, etc.

This schema was inspired highly by work done by @tetron on a (cwl specific) implementation of this API.

@geoffjentry As far as I can tell there is currently no workflow execution schema (only a task execution schema and a tool registry schema).

src/main/resources/swagger/ga4gh-workflow-execution.yaml
```diff
@@ -0,0 +1,289 @@
...
+  license:
+    name: Apache 2.0
+    url: http://www.apache.org/licenses/LICENSE-2.0.html
+paths:
+  /jobs:
+    post:
+      summary: submit a new job
```
geoffjentry: `job` likely doesn't mean what you intend for it to mean. Please refer to https://github.com/ga4gh/workflow-execution-schemas/blob/master/Definitions.md

nielsdrost:
I've removed references to CWL (except in the example used)

src/main/resources/swagger/ga4gh-workflow-execution.yaml
geoffjentry: Gotcha. The reason for that definition doc was that everybody in the batch-y type of space uses the same large handful of words to mean different things (e.g. in my world `job` is what doc describes as a `TaskInstance`). 

Prior to proceeding with merging we should discuss what the appropriate term is for this concept. I don't like using `job` (it was purposefully avoided previously) but following the previous pattern we'd end up with `WorkflowInstance` which seems unwieldy at best.

nielsdrost:
Closing, superseeded by #9

nielsdrost:
Closed pull request to not confuse people, and lack of recent activity here. Will keep following the ongoing discussion. :-)

briandoconnor:
Thanks @nielsdrost, glad to see activity picking up on this schema.  It would be great to get your opinions, feedback, and ideas on #9

PR #5, opened 2016-12-05
Linking in task execution schema

kellrott:


kellrott:
This is a draft proposal for the workflow execution schema that uses protobuf and extends on the task execution schema.

geoffjentry:
@kellrott Do you think it'd be worth discussing this on a call or not really?

kellrott:
Probably only after @tetron has looked at it

tetron:
I started playing with this in the protobuf-pa branch.

tetron:
Take a look at PR #8

PR #9, opened 2017-02-26
Feature/protobuf bdo

briandoconnor:
Need to work on the conflicts before merge.  But wanted to get this pull request open so @geoffjentry and @tetron can comment (and anyone else interested of course!)


briandoconnor:
Issues to discuss:
* Camel caps, underscore with mixed case, underscore with lowercase (I prefer the latter)... need to be consistent between TES and WES and within
* Design choice of punting on file inputs/outputs... is this going to work?
* Do we need some way to advertise supported file protocols e.g. s3://?  What about way to pass credentials?
* I added "metadata" hash to workflow submission since I thought it would handy.  Does TES want to do this as well?  I see metadata hash in their workflow status response.
* I'm using several items directly from the TES proto3... for example ga4gh_task_execJobLog.  Should these be renamed to remove "_task_"?
* WES currently has an outputs section, which to me is strange because I think we should either just put everything in the JSON param for the workflow *or* we should be explicit about inputs and outputs in WES.  Question is, how are they related to each other?  If I put URLs in the input, then something will have to stage them.  Then when I write outputs something will have to find them locally, then push them to a final destination.  We need to discuss file provisioning and see how much of this should and should not be exposed with WES.
* Should we have /wes/ in the beginning of the URL?  What about /tes/ for that service too?

briandoconnor:
@tetron and @geoffjentry I merged master into this branch

proto/workflow_execution.proto
```diff
@@ -8,58 +8,118 @@ import "google/api/annotations.proto";
...
+//workflow request object
 message WorkflowRequest {
-  string workflow_url = 1;
-  google.protobuf.Struct input = 2;
+  //OPTIONAL
+  //user name of the workflow
+  string name = 1;
+  //OPTIONAL
+  //parameter for execution engine to define/store group information
+  string project_id = 2;
```
geoffjentry: What are some examples of this? At the very least I think this needs to be well documented. What would be the motivating use case for someone taking advantage of this?

proto/workflow_execution.proto
```diff
@@ -8,58 +8,118 @@ import "google/api/annotations.proto";
...
+  //OPTIONAL
+  //free text description of task
+  string description = 3;
```
geoffjentry: In general I prefer these sorts of not-used-by-the-engine bits of arbitrary metadata to not be in the APIs (see my complaints in [a TES PR](https://github.com/ga4gh/task-execution-schemas/pull/21) for another example).

I feel that they muddy the waters and when one is trying to make a generic API there's a serious risk of overfitting one's vision of how it might be used. Further, it seems like a job for the calling infrastructure. 

If there is a good use case for description or metadata (or even the workflow name, although i feel a lot less strongly about that one) in a manner which is being directly acted upon by a workflow engine I'll retract my stance, but I'm not keen on encoding a "here, hold my stuff" portion to the API.

geoffjentry:
@briandoconnor Thanks! A few thoughts on your bullet points:

- re formatting, I don't care what the decision is but agree it should be uniform
- re file inputs. If you're saying what I think you're saying I like the idea as a first pass to be super unopinionated. "Hey WES server, here's a (CWL|WDL) and some related stuff. I'm assuming you know what to do with the related stuff now that you know if it is CWL or WDL".
- re Protocols: Despite what I just said you raise a good point although I think that might start opening a can of worms leading us down the non-KISS path
- re Credentials: None of these APIs are usable by most people w/o this, but we haven't let that stop us before :)
- re Metadata: I made my comments in the code changes prior to seeing this. I'm against adding random bits of arbitrary info in the "here, hold my stuff" capacity. If there are use cases where an implementing engine would be able to act on these fields I'll retract my stance
- re TES proto3: agree
- re `wes` in URL: Which URL?

proto/workflow_execution.proto
briandoconnor: This is taken from TES.  Let's discuss with Adam and Kyle.

geoffjentry: I probably complained about it there too :)

proto/workflow_execution.proto
briandoconnor: I understand what you mean.  I would not have put description and project in but I saw them in TES and I think we should be consistent with that API.  I'm personally OK with having arbitrary metadata being carried along since that's super useful for other systems that may need a place to store additional tracking data.  But I can see it being abused by engines that start *requiring* extra content in the metadata hash.

geoffjentry: Right. 

I'd be ok w/ these things if they were optional for the *implementer* but that doesn't really work as smoothly as being optional for the implementation. 

I can totally see why these bits of info are useful in the larger picture, but I don't think APIs like this should be attaching loosely connected stuff because then invariably what will happen is that there'll be an expectation that they're used somehow. If there's really some sort of structure on those data to me it implies another API/service.

To give a slightly different example than your last statement, I could see going down a path where there's an expectation that an engine is storing this data and one can retrieve it. No. That can't be the case. That's something else.

kellrott:
As written right now, the WorkflowStatus record only has a single ga4gh_task_exec.JobLog entry, so it only stores a single job. This should be an array.

geoffjentry:
@kellrott I think (I hope) that was unintentinoal. this should be returning (IMO) workflow status, not individual job statuses

kellrott:
Having the workflow_type as an enumeration seems a little limited. It might be better to have that as a string, so a server could extend the spec with additional workflow types. 
It might also make sense to have a 'GetServiceInfo' method similar to what TES has. The return on the service info could list the different workflow/versions that the service provides.

kellrott:
@geoffjentry I would hope that the API provides a mechanism to introspect the individual jobs that were made as a part of the workflow.

geoffjentry:
@kellrott sure but not as part of the simple status endpoint, although perhpas we're not talking about the same thing

tetron:
@briandoconnor I think this is the only WES implementation right now, do you want to update it to work with the proposed API?

https://github.com/common-workflow-language/cwltool-service/tree/ga4gh-wes

tetron:
> WES currently has an outputs section, which to me is strange because I think we should either just put everything in the JSON param for the workflow or we should be explicit about inputs and outputs in WES. Question is, how are they related to each other? If I put URLs in the input, then something will have to stage them. Then when I write outputs something will have to find them locally, then push them to a final destination. We need to discuss file provisioning and see how much of this should and should not be exposed with WES.

In my version this was intended to be the output object you get from a CWL workflow.  It is null until the workflow completes, and then it holds the result.

denis-yuen:
@kellrott  re:  "workflow_type as an enumeration seems a little limited." I agree, it also seems odd having a unknown type as an actual type. It seems oddly self-referential or something. 

@briandoconnor  re:  "Should we have /wes/ in the beginning of the URL? What about /tes/ for that service too?" tool registry schema uses " /api/ga4gh/v1" as a prefix if we want to be consistent. https://github.com/ga4gh/tool-registry-schemas/blob/develop/src/main/resources/swagger/ga4gh-tool-discovery.yaml#L9

geoffjentry:
@denis-yuen Definitely agree about the "UNKNOWN" type. Initially I was very pro-enumeration but thinking about it over the last day or so I'm definitely coming around to Kyle's suggestion. The gotcha there is enforcing consistency -> i.e. "cwl" vs "CWL" vs "commonwl", etc.

Re the api prefixes that's a good point. In retrospect I'm not a huge fan of just saying `ga4gh` but it's also what is already there.

proto/workflow_execution.proto
briandoconnor: OK, I removed name and project_Id (and description) since all of these can be modeled as key-values.

proto/workflow_execution.proto
briandoconnor: Based on the call last week (20170306) it sounds like we got the group consensus on 1) removing these name, description, project_id fields from the API spec and 2) supporting key_values that can be used to store this type of info.

briandoconnor:
Hmmm... the ability to comment inline as a reply seems limited to comments on lines of code.  Am I missing something?  Really strange to not be able to reply to comments...

So I'll just respond to the above here and make edits to the schema as I go:

1) @kellrott let's discuss the status output tomorrow on the call. I can see this working in a few different ways and I think we need to decide how we want to handle this as a group.  I'm thinking WES is really focused on reporting the status of the workflow.  *But* the moment you report a failure the first question is going to be "what job in the workflow failed and what's it's log?"

2) @geoffjentry @denis-yuen @tetron @kellrott enumeration, I agree with moving this to a string although it opens up the "CWL" vs "cwl" etc issue.  I think the spec should define official spellings with a GetServiceInfo endpoint like Kyle recommended.  And if a given system supports something not covered by the spec, then that's fine.

3) @kellrott is it sufficient to simply report back the original CWL/WDL to get information about the jobs in a workflow?  Is that the minimum viable here?  Or are you specifically referring to status of each job rather than just information about what jobs a workflow contains?

4) For consistent endpoints, I vote for:

    /ga4gh/trs/v1 -> the tool registry service
    /ga4gh/tes/v1 -> the task execution service
    /ga4gh/wes/v1 -> the workflow execution service
    /ga4gh/dos/v1 -> the data object service

Can we try to lock this down on the next call?  Does this gel with the other read/variant/etc data endpoints?

5) @tetron yes, would love to see https://github.com/common-workflow-language/cwltool-service/tree/ga4gh-wes updated to this API.  Maybe after the call tomorrow when we incorporate feedback and hopefully get closer to consensus on a 1.0.0?

6) @tetron  I think we need to talk about how to handle outputs.  My gut feeling is the output is that JSON that comes out of a cwltool run or the equivalent.  But I think TES has something to say about this.

denis-yuen:
@briandoconnor  re: 4) we made up that convention so unless there's a convenient coincidence, it probably doesn't match anything outside our task team. 

Taking a quick look, Beacon uses URLs with a prefix of api like `https://beacon-network.org/api/chromosomes` https://beacon-network.org/#/developers/api/beacon-network#bn-chromosomes
The GA4GH (read server) API looks like it just hangs off of the root ` curl --data '{}' --header 'Content-Type: application/json' \
http://localhost:8000/datasets/search | jq .` https://ga4gh-reference-implementation.readthedocs.io/en/latest/demo.html#demo

README.md
```diff
@@ -1,7 +1,7 @@
 ![ga4gh logo](http://genomicsandhealth.org/files/logo_ga.png)
 
-Schemas for the Workflow Execution API
-======================================
+Schemas for the Workflow Execution Service (WES) API
```
denis-yuen: Test

swagger/proto/workflow_execution.swagger.json
```diff
@@ -147,9 +203,36 @@
...
+      "type": "object",
+      "properties": {
+        "container": {
+          "type": "integer",
+          "format": "int32",
+          "title": "REQUIRED \nExposed port on container"
+        },
+        "host": {
+          "type": "integer",
+          "format": "int32",
```
denis-yuen: Host here is an integer, in a different location it is a string.

briandoconnor:
post on DWG from Matthew:

Dear all,

Thank you for all the useful feedback on the Use Cases document last month. As the next step in the refactoring and standardization for the API we've produced a high level design principles document.

https://github.com/Ensembl/schemas/blob/master/doc/source/specifications.rst

We'd very much welcome feedback on this document as well. The next step after this is refined will be creating some demonstration endpoints for Ensembl data using these principles.

Thanks again for all the comments

briandoconnor:
Comments to discuss on the 20170508 group call:

* general
    * renamed the base URL, do folks like this?  Should we use it across our APIs?
    * switching to snake_case for better consistency with TES
    * moved the state enum into here and a few others, eliminated dependency on TES to avoid funky naming conventions... really should move common elements to GA4GH schema repo and both TES and WES should use it.
    * avoiding how authentication/authorization works, how credentials for object stores are handed around... at least for now deferring  to the implementation
* service-info
    * service info endpoint renamed to be consistent with TES
    * added system state count (e.g. how many workflows are in each state)
    * added listing of file protocols (s3, sftp, etc) supported, we're going to punt on handing credentials around.
* GET /workflows
    * no filtering, may need to revisit later since I think this will be a burden on calling systems
    * made a note about not guarantees about mutability of listings during paging
* POST /workflows
    * made a note about the lack of mechanism to hand off object store credentials for now.
* GET /workflows/{workflow_id}
    * add log for the whole workflow and for individual steps (expecting the individual steps to be optional)
    * simplified logs and outputs to remove some of the details of TES (maybe debatable?)
    * output may include parameter and directory types in addition to output file URLs

briandoconnor:
From the call on 20170508:

* defer to implementors on authentication/authorization and on passing credentials for object storage access (can use key-values to pass extra info).  Later we'll re-evaluate after we learn more
* add a workflow_descriptor_url which can be used in place of workflow_descriptor in the POST

briandoconnor:
Seems like, based on the call, this is in a reasonable state and it should be merged and released as a 0.1

PR #11, opened 2017-06-14
0.1.0 cwl-runner-wes tweaks

tetron:
Schema feedback based on cwl-runner-wes implementation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,23 +20,19 @@ interoperability.  The Containers & Workflows working group is an informal, mult
 
 What is WES?
 ============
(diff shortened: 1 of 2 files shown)


tetron:
Proposed changes:

* `workflow_params` and `outputs` are structured JSON type instead of a string (for workflow_params) and a list of files (for outputs).  This gives symmetry between inputs and outputs.
* Add `workflow_url` to run request as an alternative to `workflow_descriptor`

https://github.com/common-workflow-language/cwltool-service/pull/1 is updated for the schema in this PR.

tetron:
Additional discussion points:

* Including all task_logs in the workflow log request may present a scaling problem when there are 100s-1000s of tasks
* Providing a state notification callback URL (eg a webhook)
* Passing through authentication (user role)

kislyuk:
> Including all task_logs in the workflow log request may present a scaling problem when there are 100s-1000s of tasks

I agree and I think it would be best to remove logs or log snippets from this endpoint altogether. Logs should be handled through a separate API endpoint that can paginate to fit them, and potentially support long-polling or other streaming mechanisms.

PR #14, opened 2018-01-10
Updates, HCA feedback

david4096:
Meant to clean up the repo and make some schema changes to accommodate recent feedback from Marcus at HCA.

* Removes dead code
* Begins alignment with TES (thanks @kellrott)
* Improves comments and protobuf styling

Work in progress


thejmazz:
with regards to changing `workflow_params` from `google.protobuf.Struct` to `string`:

Is there an example showing how "JSON you give a WDL executor is not typed, just strings", from [here](https://software.broadinstitute.org/wdl/documentation/spec#specifying-workflow-inputs-in-json), it looks like JSON?

If it is set to string, would then stringified JSON be compatible?

tetron:
Changing workflow_parameters from a structured record to a raw string with arbitrary format that has to be parsed is a step backwards in interoperability.  This increases the friction involved in connecting a WES endpoint to other systems, as it imposes additional free text input/output format munging.  And if the practical outcome is "everyone should use stringified JSON" then why make it a string field in the first place?

david4096:
Hi @thejmazz and @tetron! Thanks!!!! It looks like the input can be YAML or JSON, and so using an open string allows both cases.

https://github.com/openwdl/wdl/blob/master/versions/draft-2/SPEC.md#specifying-workflow-inputs-in-json

@briandoconnor @mckinsel any opinions regarding this field? I'll roll it back for now!

mckinsel:
Hey everybody, thanks for the draft and comments.

I believe this suggestion started with my concern that valid values for `workflow_params` will vary by platform. The most pressing instance of this for me is that WDL inputs are not explicitly typed, and certain required type conversions can't be implemented easily in some platforms. After some discussion of the issue, I think we decided that this field needed to be maximally flexible.

I don't really feel too strongly about it, but I'm doubtful that there's much of an interoperability difference between an unconstrained string and unconstrained JSON.

Also, I see this change from a struct to a string in the proto file, but the swagger json already has `workflow_params` as a string. Is it possible for the two WES definitions to get out of sync, or am I looking in the wrong spot?

david4096:
@mckinsel sounds like a [version issue](https://github.com/ga4gh/workflow-execution-schemas/commit/e0cad9ba3e5fdaa823d03526302fc18c8f48675e), this PR will add the generated swagger to help avoid this. Sorry!

proto/workflow_execution_service.proto
```diff
@@ -0,0 +1,237 @@
...
+  // Sample of stdout (not guaranteed to be entire log)
+  string stdout = 5;
+  // Sample of stderr (not guaranteed to be entire log)
+  string stderr = 6;
+  // Exit code of the program
+  int32  exitCode = 7;
+}
+
+// Enumeration of states for a given workflow request
+enum state {
```
kellrott: This is basically just recasting the TES state declaration ( https://github.com/ga4gh/task-execution-schemas/blob/master/task_execution.proto#L355 ) but with a different order for the values ( Initializing = 8; vs  INITIALIZING = 2; ) and the string equivalents will be 'Initializing' vs 'INITIALIZING'.

proto/workflow_execution_service.proto
```diff
@@ -0,0 +1,237 @@
...
+  // The workflow descriptor type, must be "CWL" or "WDL" currently (or another alternative supported by this WES instance)
+  string workflow_type = 3;
+
+  // REQUIRED
+  // The workflow descriptor type version, must be one supported by this WES instance
+  string workflow_type_version = 4;
+
+  // OPTIONAL
+  // A key-value map of arbitrary metadata outside the scope of the workflow_params but useful to track with this workflow request
+  map<string,string> key_values = 5;
```
kellrott: Same concept as the TES `tags` ( https://github.com/ga4gh/task-execution-schemas/blob/master/task_execution.proto#L66 ) but with a different name

proto/workflow_execution_service.proto
```diff
@@ -0,0 +1,237 @@
...
+  // simply contains an array of one or more version strings
+  map<string, workflow_type_version> workflow_type_versions = 1;
+
+  // The version(s) of the WES schema supported by this service
+  repeated string supported_wes_versions = 2;
+
+  // The filesystem protocols supported by this service, currently these may include common
+  // protocols such as 'http', 'https', 'sftp', 's3', 'gs', 'file', 'synapse', or others as
+  // supported by this service.
+  repeated string supported_filesystem_protocols = 3;
```
kellrott: We should have a discussion about syncing this concept with the one used in TES: https://github.com/ga4gh/task-execution-schemas/blob/master/task_execution.proto#L527

proto/workflow_execution_service.proto
```diff
@@ -0,0 +1,237 @@
...
+    option (google.api.http) = {
+      get: "/ga4gh/wes/v1/workflows"
+    };
+  }
+
+}
+
+// To execute a workflow, send a workflow request including all the details needed to begin downloading
+// and executing a given workflow.
+message workflow_request {
```
kellrott: Protobuf style guide suggests camel case for message names ( https://developers.google.com/protocol-buffers/docs/style )

proto/workflow_execution_service.proto
david4096: I created an issue https://github.com/ga4gh/workflow-execution-schemas/issues/15

david4096:
Hi Everyone! I removed the work in progress title and am hoping to merge this today! It's mostly cosmetic changes that remove some dead code, improve style, and align with other GA4GH efforts. The repository should be a little easier to work with now (fewer directions, requirements).

briandoconnor:
@tetron I like the idea of having a standardized input JSON regardless of underlying execution system.  But for right now, given that both Cromwell and CWLtool use different JSON param formats it means anyone implementing WES will need to translate from WES to the execution-specific format.  It seems like we should be able to, as a community, come up with a common JSON parameterization format for CWL and WDL.  Do you think that's possible?  If we had that, and it happened to be the same format as the WES param JSON that would be the best case scenario.

briandoconnor:
@david4096 do you have this swagger hosted somewhere?  Would be great to look at this API through that interface

briandoconnor:
FYI, @david4096 provided: https://david4096.github.io/workflow-execution-schemas/

briandoconnor:
Some feedback @david4096 :

* general
    * I think we need better details in the endpoint descriptions (not a blocker for your pull request, just need this before release)
* GET /ga4gh/wes/v1/service-info
    * engine_versions -> workflow_engine_versions: Just to make it a little more clear
    * auth_instructions_url: add this so a site can provide a URL a user goes to to log in and generate their auth token... I don't think we do anything more than this right now with auth (other than having certain endpoints that need an authorized bearer token)
    * I thought the feedback from the HCA Driver was they wanted a named key-value reporting what extra params the workflow engine would take, their types, and default values?  Is the approach right now to overload the tags with this?
* POST /ga4gh/wes/v1/workflows
    * I still would prefer workflow_params be a string until we can actually give people a JSON schema used across workflow engines.  Either string or JSON I can't parse it at this point since WDL and CWL use different schemas and it's not like we're proposing any structure.  I'd prefer a 1.1 version of WES to provide a common JSON format regardless workflow engine (like we do with the output for example).
    * are we using tags to pass extra, optional "hints" to the workflow engine?  See comment above in service-info endpoint.
    * workflow_descriptor: we said this would, optionally, accept base64 encoded zip files for systems that support it.  What happens if this is a zip?  How will the execution system know which WDL/CWL inside the zip is to be executed?  In that case do you need to pass in the path that you want to execute via the workflow_url param?  Messy but meets the needs of HCA and Marcus

I think that's it.  Need responses on these ASAP since we want to cut a release soon to have something stable to share with the larger Commons effort.  I think you're really close though!  Thanks David and everyone else that's worked hard to push this forward!!

david4096:
Thanks @briandoconnor I'll incorporate what I can from your comments and push the rest out to issues!

david4096:
Hi everyone! I updated the [OpenAPI description here](https://david4096.github.io/workflow-execution-schemas/#/WorkflowExecutionService/RunWorkflow) to include @briandoconnor's comments. Thanks!

briandoconnor:
Cool @david4096 !  I'll take a look when I get home tonight and merge

david4096:
FYI I left the workflow_params as a struct, which is a `oneof` an Object, ListValue, string, etc. which should support the needs of HCA.

PR #24, opened 2018-03-02
Remove proto, add error responses, setup.py

### LINKED ISSUE #17: Error handling
The response messages for a bad requests, not found resources, etc. are not described in the schema.

Can this be done directly in protobuf? Should this be done via modifying the OpenAPI description directly?

david4096:
Will remove the protocol buffers bindings and make some improvements to the OpenAPI description.

`WorkflowDesc` was a duplicate of `WorkflowStatus` so merged in favor of WorkflowStatus.

Error Responses were added.

The `ProtobufStruct` objects have been replaced with arbitrary JSON objects.

A client based on bravado and a stub server in connexion were added.

Close #17


briandoconnor:
No feedback and on the call got thumbs up to move away from proto... so merging...

PR #29, opened 2018-05-21
Add contributing doc, fixes

david4096:
Adds a basic contributing doc based on ga4gh-schemas.

Fixes a problem with the schemas where the "titles" were set when "description" was desired. This causes problems in clients downstream.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,215 @@
+How to contribute to the GA4GH Schemas
+======================================
+
+Thank you for taking the time to contribute. We appreciate it!
+
+The GA4GH schemas define an API for sharing genomic and phenotypic data.
+
+There are two ways to contribute to the effort - via issues, which are
+used for discussion, and pull requests, which are concrete proposals of
+change.
+
+Issues
+======
+
+The project's [Issues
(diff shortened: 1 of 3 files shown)


jaeddy:
+1

rishidev:
At the moment this doesn't capture the procedure we outlined for mandatory +1s from named people and veto power from Work Stream leads. There is also some GA4GH related terminology to tidy up - we don't have DWG and Task Teams any more. The Contributing.MD doc itself forms a good basis to work with though, so we could accept and refine. However I'm not comfortable with resolving the conflicts above so @jaeddy - could you advise please - would it be better to accept this PR then update with a follow up or to make a new PR?

geoffjentry:
@rishidev Do you have the canonical list of current thumbs-up-givers? There's myself, @tetron @jaeddy @mckinsel and who else?

rishidev:
@geoffjentry That is the list of four required thumbs up givers.

briandoconnor:
What's the difference between title and description @david4096 ?

david4096:
@rishidev please open another PR based on those details, this was taken from the other ga4gh-schemas feel free to modify.

@briandoconnor the title's error is from the protobuf conversion and shouldn't be there. It causes problems when generating downstream clients as they try to make accessors called "Your very long title". We'll fix that in a separate PR.

PR #30, opened 2018-06-05
Clarify logs (expected to be URLs) and paging (use page_token).

tetron:
Also clarify that timestamps should be ISO 8601.

Follow up from Toronto meeting.

Corresponding implementation here https://github.com/common-workflow-language/workflow-service/pull/21

paging @geoffjentry @mckinsel @jaeddy @briandoconnor @dglazer

### DIFF
--- a/openapi/workflow_execution_service.swagger.yaml
+++ b/openapi/workflow_execution_service.swagger.yaml
@@ -44,10 +44,13 @@ paths:
   /workflows:
     get:
       summary: |-
-        List submitted workflows.  The ordering of this list is implementation dependent.
-        There is no guarantee of live updates as the user traverses the pages, the behavior should be
-        decided (and documented) by each implementation.
-        To monitor a given execution, use GetWorkflowStatus or GetWorkflowLog.
+        List submitted workflows.  This should be provided in a stable
+        ordering, however the ordering of this list is implementation
+        dependent.  When paging through the list, the client should
+        not make assumptions about live updates, but should assume the
+        contents of the list reflect the workflow list at the moment
+        that the first page is requested.  To monitor a specific
+        workflow run, use GetWorkflowStatus or GetWorkflowLog.
       x-swagger-router-controller: ga4gh.wes.server
       operationId: ListWorkflows
       responses:
@@ -75,8 +78,14 @@ paths:
         - name: page_size
           description: |-
             OPTIONAL
-            The preferred number of workflow submissions to return in a page.
-            The actual number of items returned is implementation dependent.
+            The preferred number of workflow runs to return in a page.
+            If not provided, the implementation should use a default page size.
+            The implementation must not return more items
+            than "page_size", but it may return fewer.  Clients should
+            not assume that if fewer than "page_size" items is
+            returned that all items have been returned.  The
+            availability of additional pages is indicated by the value
+            of "next_page_token" in the response.
           in: query
           required: false
           type: integer
@@ -92,7 +101,8 @@ paths:
         - name: tag_search
           description: |-
             OPTIONAL
-            Only return workflow submissions where this key is present in "tags" (regardless of tag value).
+            For each key, if the key's value is empty string then match workflows that are tagged with
+            this key regardless of value.
           in: query
           required: false
           type: string
--- a/openapi/workflow_execution_service.swagger.yaml
+++ b/openapi/workflow_execution_service.swagger.yaml
@@ -44,7 +44,7 @@ paths:
   /workflows:
     get:
       summary: |-
-        List the workflows, this endpoint will list the workflows in order of oldest to newest.
+        List submitted workflows.  The ordering of this list is implementation dependent.
         There is no guarantee of live updates as the user traverses the pages, the behavior should be
         decided (and documented) by each implementation.
         To monitor a given execution, use GetWorkflowStatus or GetWorkflowLog.
@@ -75,7 +75,8 @@ paths:
         - name: page_size
           description: |-
             OPTIONAL
-            Number of workflows to return in a page.
+            The preferred number of workflow submissions to return in a page.
+            The actual number of items returned is implementation dependent.
           in: query
           required: false
           type: integer
@@ -83,7 +84,7 @@ paths:
         - name: page_token
           description: |-
             OPTIONAL
-            Token to use to indicate where to start getting results. If unspecified, returns the first
+            Token to use to indicate where to start getting results. If unspecified, return the first
             page of results.
           in: query
           required: false
@@ -91,8 +92,7 @@ paths:
         - name: tag_search
           description: |-
             OPTIONAL
-            For each key, if the key's value is empty string then match workflows that are tagged with
-            this key regardless of value.
+            Only return workflow submissions where this key is present in "tags" (regardless of tag value).
           in: query
           required: false
           type: string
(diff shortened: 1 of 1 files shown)


openapi/workflow_execution_service.swagger.yaml
```diff
@@ -75,24 +75,24 @@ paths:
...
             page of results.
           in: query
           required: false
           type: string
         - name: tag_search
           description: |-
             OPTIONAL
-            For each key, if the key's value is empty string then match workflows that are tagged with
-            this key regardless of value.
+            Only return workflow submissions where this key is present in "tags" (regardless of tag value).
```
dglazer: making sure I understand:
a) we let you search for "key K exists", but don't provide any way to search for "key K has value V"
b) we let you search for a single key, but don't provide any way to search for "key J and/or K exist"
Is that right? If so, it seems like a reasonable place to start, but I could be talked into doing more or less depending on what our initial driver projects need.

tetron: I don't have strong feelings about this, I was just trying to reword it to make it clearer what behavior was being specified.

geoffjentry: I don't care strongly either way but without claiming to speak for @mckinsel - what we've seen from the group he works with using our stuff is they want both K and K/V pair as K is sometimes used as a straight tag and K/V pair is used as a, well, K/V pair.

geoffjentry:
:+1: modulo if any changes are requested by drivers on the K/V issue

openapi/workflow_execution_service.swagger.yaml
```diff
@@ -75,24 +75,24 @@ paths:
         - name: page_size
           description: |-
             OPTIONAL
-            Number of workflows to return in a page.
+            The preferred number of workflow submissions to return in a page.
+            The actual number of items returned is implementation dependent.
```
ddietterich: I think it is important for clients to know the maximum number of workflows they will need to handle in a response. That affects structure allocation. The number could be fewer, such as at the end of the listing, but shouldn't be more.

openapi/workflow_execution_service.swagger.yaml
```diff
@@ -44,7 +44,7 @@ paths:
   /workflows:
     get:
       summary: |-
-        List the workflows, this endpoint will list the workflows in order of oldest to newest.
+        List submitted workflows.  The ordering of this list is implementation dependent.
```
ddietterich: The implementation really has to provide some kind of ordering; maybe implementation-specific. Otherwise, the semantic below of returning to the "first" result has no meaning.

openapi/workflow_execution_service.swagger.yaml
ddietterich: If we need to support both search for "K exists" and "K with value" exists, then I think tag_search needs to allow specification of that. For example, tag_search is structure that contains "key" field (required if tag_search is specified) and "value" field (optional; if not present we only check key existence.

From my naive point of view, this seems a bit ornate. I'd like to understand the use case for "key + value" and why a simple set of string tags is not sufficient.

jaeddy: It'd also be helpful to clarify how these tags are related (if at all) to the "tags" field of [`WorkflowRequest`](https://github.com/ga4gh/workflow-execution-service-schemas/blob/27b1701206d7c7171d0dc77eb7c4d4a35c3e6953/openapi/workflow_execution_service.swagger.yaml#L435-L487):

```
    tags:
        type: object
        additionalProperties:
          type: string
        title: |-
          OPTIONAL
          A key-value map of arbitrary metadata outside the scope of the workflow_params but useful to track with this workflow request
```

geoffjentry: Meta comment: user kbergin made a comment and it’s been listed as pending for a few hours. Who can unblock this?

geoffjentry: @tetron do you have the ability to approve the comment from @kbergin ?

tetron: https://help.github.com/articles/reviewing-proposed-changes-in-a-pull-request/

> Before you submit your review, your line comments are pending and only visible to you.

That might be what's going on?

geoffjentry: Oh I bet you’re right. I was going to say, I’d never seen a setting requiring comment approvals.

kbergin: So sorry for the meta-comment confusion!

tetron:
I'm going to pull the text changes related to tags (I was only trying to clarify what was there and not propose new behavior, but clearly we need to talk about it more) and we can discuss it separately.

tetron:
Updated to remove text changes related tags and to reflect comments from @ddietterich

openapi/workflow_execution_service.swagger.yaml
```diff
@@ -75,15 +78,22 @@ paths:
...
           type: integer
           format: int64
         - name: page_token
           description: |-
             OPTIONAL
-            Token to use to indicate where to start getting results. If unspecified, returns the first
+            Token to use to indicate where to start getting results. If unspecified, return the first
             page of results.
           in: query
           required: false
```
jaeddy: Does "page_token" need a type? I assume string (to match "next_page_token").

openapi/workflow_execution_service.swagger.yaml
```diff
@@ -75,15 +78,22 @@ paths:
...
           description: |-
             OPTIONAL
-            Number of workflows to return in a page.
+            The preferred number of workflow runs to return in a page.
+            If not provided, the implementation should use a default page size.
+            The implementation must not return more items
+            than "page_size", but it may return fewer.  Clients should
+            not assume that if fewer than "page_size" items is
+            returned that all items have been returned.  The
+            availability of additional pages is indicated by the value
```
jaeddy: Is there a "default" behavior for returning workflows when neither "page_size" or "page_token" are specified? Does the WES implementation itself specify a default page size?

Also, if 'Clients should not assume that if fewer than "page_size" items is returned that all items have been returned,' is the "system_state_counts" property in `ServiceInfo` a more reliable way to check the total count of submitted workflows?

openapi/workflow_execution_service.swagger.yaml
```diff
@@ -264,16 +274,26 @@ definitions:
...
+        title: When the command started executing, in ISO 8601 format "%Y-%m-%dT%H:%M:%SZ"
       end_time:
         type: string
-        title: When the command completed
+        title: When the command stopped executing (completed, failed, or cancelled), in ISO 8601 format "%Y-%m-%dT%H:%M:%SZ"
       stdout:
         type: string
-        title: Sample of stdout (not guaranteed to be entire log)
+        title: |-
+          A URL to retrieve standard output logs of the workflow run or
```
jaeddy: Might be useful to provide client some indication of what protocol is required to access log URLs (e.g., http, gs, s3, etc.). I'm thinking that `supported_filesystem_protocols` in `ServiceInfo` should be a map instead of an array (e.g., `{"workflows": ["http", "file"], "tasks": ["gs"], "logs": ["gs"]}`) — but I'll open a separate issue to discuss.

openapi/workflow_execution_service.swagger.yaml
tetron: It does have `type: string`, I think you just missed it (it is on line 100, which is just below the context provided for this line comment).

openapi/workflow_execution_service.swagger.yaml
tetron: I guess it isn't clear, the intended default behavior is to start on the first page and the page size is arbitrary.

tetron:
In terms of process, how do we wrap this up?  I see approval from @geoffjentry and @jaeddy so I guess @mckinsel is supposed to also weigh in?

dglazer:
Yes in terms of process -- Marcus, can you either say +1 or let us know
your concerns? (Or choose to defer to the other reviewers, which is also
fine.)

PR #31, opened 2018-06-05
Propose spec changes for multipart upload.

tetron:
Add text describing multipart upload.

Remove workflow_descriptor because it is superceded by multipart
upload.

Follow up from Toronto meeting.

Corresponding proof of concept implementation here https://github.com/common-workflow-language/workflow-service/pull/22

paging @geoffjentry @mckinsel @jaeddy @briandoconnor @dglazer

### DIFF
--- a/openapi/workflow_execution_service.swagger.yaml
+++ b/openapi/workflow_execution_service.swagger.yaml
@@ -403,7 +403,7 @@ definitions:
         description: |-
           A web page URL with information about how to get an
           authorization token necessary to use a specific endpoint.
-     contact_info:
+      contact_info:
         type: string
         description: |-
           An email address or web page URL with contact information
@@ -480,7 +480,7 @@ definitions:
           A token which may be supplied as "page_token" in workflow run list request to get the next page
           of results.  An empty string indicates there are no more items to return.
     description: The service will return a workflow_run_list_response when receiving a successful workflow_run_list_request.
-  WorkflowLog:
+  WorkflowRunLog:
     type: object
     properties:
       run_id:
--- a/openapi/workflow_execution_service.swagger.yaml
+++ b/openapi/workflow_execution_service.swagger.yaml
@@ -103,32 +103,35 @@ paths:
         Run a workflow.  This endpoint creates a new workflow run and
         returns the workflow ID to monitor its progress.
 
-        The workflow_params JSON object specifies input parameters,
+        The request may upload files that are required to execute the
+        workflow identified as `workflow_attachment`.  The parts
+        supplied in `workflow_attachment` may include the primary
+        workflow, tools imported by the workflow, other files
+        referenced by the workflow, or files which are part of the
+        input.  The implementation should stage these files to a
+        temporary directory and execute the workflow from there.
+        These parts must have a Content-Disposition header with a
+        "filename" provided for each part.  Filenames may include
+        subdirectories, but must not include references to parent
+        directories with '..', implementations should guard against
+        maliciously constructed filenames.
+
+        The `workflow_url` is either an absolute URL to a workflow
+        file that is accessible by the WES endpoint, or a relative URL
+        corresponding to one of the files attached using
+        `workflow_attachment`.
+
+        The `workflow_params` JSON object specifies input parameters,
         such as input files.  The exact format of the JSON object
         depends on the conventions of the workflow language being
-        used.  Input files should be addressed by URL.  The WES
-        endpoint must understand and be able to access URLs supplied
-        in the input.  This is implementation specific.
+        used.  Input files should either be absolute URLs, or relative
+        URLs corresponding to files uploaded using
+        `workflow_attachment`.  The WES endpoint must understand and
+        be able to access URLs supplied in the input.  This is
+        implementation specific.
 
-        This endpoint can be used two different ways, selected by
-        Content-Type.
-
-        If Content-Type is application/json, the WorkflowRequest JSON
-        is provided in the body of the request.  The "workflow_url"
-        must refer to a workflow file by URL that is accessible by the
-        WES endpoint, in addition any files referenced in
-        workflow_params must be referenced by full URL.
-
-        If Content-Type is multipart/form-data, the request consists
-        of one or more files that are required to execute the
-        workflow.  Form parts MUST have a Content-Disposition header
-        with a "filename" provided for each part.
-
-        Parts may include the primary workflow, tools imported by the
-        workflow, other files referenced by the workflow, or files
-        which are part of the input.  The implementation should stage
-        these files to a temporary directory and execute the workflow
-        from there.
+        See documentation for WorkflowRequest for detail about other
+        fields.
 
       x-swagger-router-controller: ga4gh.wes.server
       operationId: RunWorkflow
@@ -153,12 +156,42 @@ paths:
           description: An unexpected error occurred.
           schema:
             $ref: '#/definitions/ErrorResponse'
+      consumes:
+         - multipart/form-data
       parameters:
-        - name: body
-          in: body
-          required: true
-          schema:
-            $ref: '#/definitions/WorkflowRequest'
+        - in: formData
+          name: workflow_params
+          type: string
+          format: application/json
+
+        - in: formData
+          name: workflow_type
+          type: string
+
+        - in: formData
+          name: workflow_type_version
+          type: string
+
+        - in: formData
+          name: tags
+          type: string
+          format: application/json
+
+        - in: formData
+          name: workflow_engine_parameters
+          type: string
+          format: application/json
+
+        - in: formData
+          name: workflow_url
+          type: string
+
+        - in: formData
+          name: workflow_attachment
+          type: array
+          items:
+            type: string
+            format: binary
       tags:
         - WorkflowExecutionService
   '/workflows/{workflow_id}':
@@ -480,10 +513,10 @@ definitions:
       workflow_url:
         type: string
         description: |-
-          OPTIONAL
-          The workflow CWL or WDL document, must provide either this or workflow_descriptor. When a base64 encoded gzip of
-          workflow descriptor files is offered, the `workflow_url` should be set to the relative path
-          of the main workflow descriptor.
+          REQUIRED
+          The workflow CWL or WDL document.
+          When workflow attachments files are provided, the `workflow_url` may be a relative path
+          corresponding to one of the attachments.
     description: |-
       To execute a workflow, send a workflow request including all the details needed to begin downloading
       and executing a given workflow.
--- a/openapi/workflow_execution_service.swagger.yaml
+++ b/openapi/workflow_execution_service.swagger.yaml
@@ -100,15 +100,36 @@ paths:
         - WorkflowExecutionService
     post:
       summary: |-
-        Run a workflow, this endpoint will allow you to create a new workflow request and
-        retrieve its tracking ID to monitor its progress.  An important assumption in this
-        endpoint is that the workflow_params JSON will include parameterizations along with
-        input and output files.  The latter two may be on S3, Google object storage, local filesystems,
-        etc.  This specification makes no distinction.  However, it is assumed that the submitter
-        is using URLs that this system both understands and can access. For Amazon S3, this could
-        be accomplished by given the credentials associated with a WES service access to a
-        particular bucket.  The details are important for a production system and user on-boarding
-        but outside the scope of this spec.
+        Run a workflow.  This endpoint creates a new workflow run and
+        returns the workflow ID to monitor its progress.
+
+        The workflow_params JSON object specifies input parameters,
+        such as input files.  The exact format of the JSON object
+        depends on the conventions of the workflow language being
+        used.  Input files should be addressed by URL.  The WES
+        endpoint must understand and be able to access URLs supplied
+        in the input.  This is implementation specific.
+
+        This endpoint can be used two different ways, selected by
+        Content-Type.
+
+        If Content-Type is application/json, the WorkflowRequest JSON
+        is provided in the body of the request.  The "workflow_url"
+        must refer to a workflow file by URL that is accessible by the
+        WES endpoint, in addition any files referenced in
+        workflow_params must be referenced by full URL.
+
+        If Content-Type is multipart/form-data, the request consists
+        of one or more files that are required to execute the
+        workflow.  Form parts MUST have a Content-Disposition header
+        with a "filename" provided for each part.
+
+        Parts may include the primary workflow, tools imported by the
+        workflow, other files referenced by the workflow, or files
+        which are part of the input.  The implementation should stage
+        these files to a temporary directory and execute the workflow
+        from there.
+
       x-swagger-router-controller: ga4gh.wes.server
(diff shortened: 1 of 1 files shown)


openapi/workflow_execution_service.swagger.yaml
```diff
@@ -100,15 +100,36 @@ paths:
...
+        returns the workflow ID to monitor its progress.
+
+        The workflow_params JSON object specifies input parameters,
+        such as input files.  The exact format of the JSON object
+        depends on the conventions of the workflow language being
+        used.  Input files should be addressed by URL.  The WES
+        endpoint must understand and be able to access URLs supplied
+        in the input.  This is implementation specific.
+
+        This endpoint can be used two different ways, selected by
```
geoffjentry: This is not how I understood our agreement.

I was picturing the submission request to be a multipart with:

- The JSON bundle of most things
- 0 or more workflow descriptor files

@mckinsel Since you & I have the action item to design this, I'm curious to get your read on this. Was I the outlier here?

tetron: I apologize for jumping in on your action item.  I was doing some prototyping so I wanted to write up my findings, and thought a PR would be the most productive way to kickstart the discussion.

We could specify that it only accepts multipart, but in any case one of the "parts" needs to be the json request, and the other "parts" are the files needed to run the workflow.  The part containing the json payload can either be the first item, or an item with a particular name.  I went with a particular name ("body") but it is fine either way.

Selecting on content-type makes life simpler for the submitter when they want to provide a workflow_url and not upload files.  But that case can also be accommodated with multipart with a single item ("body").

geoffjentry: Hah no I was happy you did :) It was also good because it made clear much faster that there was a difference of opinion on what was decided.

The second paragraph is exactly how I had been picturing it. 

The third paragraph works for me insofar as the JSON blob has a well specified form field when multipart

My preference would be for the second paragraph, but I tend to dislike overloading things in general so I recognize that this quite possibly just my bias showing.

tetron: Just to be completely clear, I mean choosing behavior based on Content-Type of the primary headers as to whether the body of the request is just the request json or whether it is a multipart stream, and not talking about Content-Type fields on the individual parts.

But in the interests of keeping the interface minimal, I can take out that part, we specify that:

* it always has to be multipart
* the request json is the part named "body"
* even if the submitter doesn't need to attach files, they still have to use multipart encoding on the body part

geoffjentry: We should, at a minimum, wait for Marcus to comment prior to changing the PR

mckinsel: Hey apologies for the delay in response. I like the multipart-only idea just in terms of keeping the implementation simple for now. Does it make sense to just include the other workflow request fields as form values? I'm not sure what we get out of json encoding and then multipart encoding.

Also, I don't think there's a way to specify this interface in swagger 2? @geoffjentry I think this is the issue that's led to the WorkflowParams{1..N} in the cromwell API, right?

tetron: > Hey apologies for the delay in response. I like the multipart-only idea just in terms of keeping the implementation simple for now.

I'm okay with making this simplification, but also foresee the next person looking at the spec wondering why they need to do a multipart encode when all they want to do is execute a workflow by URL.

> Does it make sense to just include the other workflow request fields as form values?   I'm not sure what we get out of json encoding and then multipart encoding.

The json payload is just one of the parts (the rest are the files).  Technically you could pass each field as part as well, but having the payload be json encoded seems more consistent with the rest of the API.

geoffjentry: @tetron "but also foresee the next person looking at the spec wondering why" FWIW when I have shown the spec to folks locally they're always wondering why the JSON blob approach is being used and it's not a multipart form altogether. I say this not as pushing for what @mckinsel suggested but rather to note that people are going to be coming to this with a variety of biases and will by nature be expecting to see what they're used to seeing.

@mckinsel Yes, there is (was?) a limitation in Swagger where you can't have arbitrarily large set of fields with the same name. That WorkflowParams{1..N} is purely a swagger thing, it's not necessary in e.g. curl, but was the only way we could get it to work at all in swagger.

mckinsel: I guess some of the non-`workflow_descriptor` fields are going to be JSON blobs either way: `workflow_params` and `workflow_engine_parameters`. So clients and implementers are doing some nested encoding/decoding in either scenario.

The swagger limitation appears to be fixed in openapi 3. You can do something like this (I think):
```yaml
content:
  multipart/form-data:
    schema:
      WorkflowRequest:
        type: object
        properties:
          workflow_descriptor:
            type: array
            items:
              type: string
              format: binary
```

tetron: If there's a preferred encoding in OpenAPI 3 then we should probably adopt that.

geoffjentry: @mckinsel On your first point I don't think that's right. IIRC those were both specified to be actual JSON objects (instead of the current "whatever the heck one wants, but we know it'll be JSON" which means they can be sub-objects in the main JSON blob. The same would be true of form fields (as we see in the Cromwell API). So I don't think either side should require escaping?

tetron:
@mckinsel @geoffjentry can we get to a consensus here?  We don't want to supportmultiple content types, it always has to be `multipart/form-data`, but I don't know whether we want fields such as `workflow_params` to be separate form parts or packed into a single json encoded form part?

geoffjentry:
@tetron I have a slight personal preference towards "just make the whole thing a multi-part" but I'm biased as that's the world I come from. I'm OK w/ that not being the case. 

I have a strong preference towards it always being some sort of multi-part, and if it's not all multipart the one which makes the most sense to me is `workflow-descriptor-X` (or whatever) and `everything-else` (or whatever)

If the final answer is one of those two then I would leave it totally up to you and @mckinsel (as the two other folks who seem to have opinions). If it's something else, I reserve the right to chime in again :)

Related: We're coming up on the quarter boundary which means I will be able to get time (and/or someone else's time) to start poking at this soon

openapi/workflow_execution_service.swagger.yaml
```diff
@@ -425,15 +446,6 @@ definitions:
...
-      workflow_descriptor:
-        type: string
-        description: |-
-          OPTIONAL
-          The workflow CWL or WDL document, must provide either this or workflow_url. By combining
-          this message with a workflow_type_version offered in ServiceInfo, one can initialize
-          CWL, WDL, or a base64 encoded gzip of the required workflow descriptors. When files must be
-          created in this way, the `workflow_url` should be set to the path of the main
-          workflow descriptor.
       workflow_params:
```
david4096: Should we do the same thing for workflow_params? @achave11

jaeddy:
After staring at the OpenAPI 3.0 spec and docs for a while (and following @mckinsel's example), I _think_ something like this might work... I could also be wrong. Regardless of specification, I'm fine with the "just make the whole thing a multi-part" approach, as long as there's a sensible way to discriminate between allowable types for `workflow_descriptor`.

```yaml
paths:
  /workflows:
    post:
      summary: Submit a workflow to run.
    requestBody:
      content:
        multipart/form-data:
          schema:
            $ref: '#/components/schemas/WorkflowRequest'

components:
  schemas:
    SimpleDescriptor:
      type: string
    ComplexDescriptor:
      type: array
      items:
        type: string
        format: binary
    WorkflowRequest:
      type: object
      properties:
        workflow_descriptor:
          oneOf:
          - $ref: '#/components/schemas/SimpleDescriptor'
          - $ref: '#/components/schemas/ComplexDescriptor'
        workflow_url:
          type: string
        workflow_params:
          $ref: '#/components/schemas/WesObject'
        workflow_type:
          type: string
        workflow_type_version:
          type: string
        tags:
          $ref: '#/components/schemas/WesObject'
        workflow_engine_parameters:
          $ref: '#/components/schemas/WesObject'
```

tetron:
A few things:

You can describe multipart file upload in swagger 2.0: https://swagger.io/docs/specification/2-0/file-upload/

It also describes a workaround for describing arbitrary file uploads as an array of "binary" format strings, this is formalized in the OpenAPI 3.0 spec (the way of describing it is almost exactly the same).

Another limitation of Swagger 2.0 is that you can't have form parts which are "object" type.  I think we can have them as strings and specify they are application/json though.

There's an online converter from Swagger 2.0 to OpenAPI 3.0:
https://openapi-converter.herokuapp.com/

Here's the problem:  the tooling we're using (at least in Python, `connexion` and `bravado`) only understands Swagger 2.0.  So if we switch to OpenAPI 3.0, what we gain in expressiveness, we lose in convenience.

My personal preference is to try and make it work with Swagger 2.0, at least for the time being.

denis-yuen:
We have the same problem with tooling in Java / Typescript for the near future.

geoffjentry:
@tetron Agree that for now it's probably worth sticking w/ 2.0 if possible for a host of practical reasons - beyond what you cited there's also the "who is volunteering to port this?" issue :)

I'd forgotten about that workaround. The case in the Cromwell API that led us to discover that functionality hole was such that the workaround wasn't useful as we explicitly wanted the Swagger API to provide the upload links, but for this purpose I think it's fine as long as the spec swagger is clear about what's happening.

Just so I'm clear, are we now agreed on the it's all multipart angle? The quarter starts next week so I'm hoping I can direct some of my group's attention this way ASAP once the switchover happens

tetron:
Updated to specify that it has to `consume: multipart/form-data` and parameters are `in: formData`, renamed `workflow_descriptor` to `workflow_attachment` and updated text.  I think this reflects the current consensus so please take a look.  The corresponding branch is updated as well: https://github.com/common-workflow-language/workflow-service/pull/22

briandoconnor:
@jaeddy is going to resolve conflicts.

Next week we call for a vote leading up to the next Monday call?

geoffjentry:
Tagging @mckinsel since Brian wasn't able to get the user ID live during the call

openapi/workflow_execution_service.swagger.yaml
```diff
@@ -110,15 +110,39 @@ paths:
...
+        temporary directory and execute the workflow from there.
+        These parts must have a Content-Disposition header with a
+        "filename" provided for each part.  Filenames may include
+        subdirectories, but must not include references to parent
+        directories with '..', implementations should guard against
+        maliciously constructed filenames.
+
+        The `workflow_url` is either an absolute URL to a workflow
+        file that is accessible by the WES endpoint, or a relative URL
+        corresponding to one of the files attached using
```
geoffjentry: @tetron Can you add an example for the latter case? I had to think about what that'd look like for a moment, and this seems like a case where a little ambiguity can lead to differences in behavior

openapi/workflow_execution_service.swagger.yaml
```diff
@@ -142,12 +166,42 @@ paths:
...
+        - in: formData
+          name: workflow_params
+          type: string
+          format: application/json
+
+        - in: formData
+          name: workflow_type
+          type: string
+
+        - in: formData
```
geoffjentry: Idle thought - what is the appropriate response when the workflow type & version don't match reality? Is there an expectation that this be checked, or is it just "run it like this and hope for the best"?

We could also consider asking implementations to auto-detect these (e.g. it was pretty trivial to add this to Cromwell, we no longer require the equivalent fields in Cromwell's submit endpoint as we just autodetect wdl vs cwl and which version it is)

briandoconnor:
@mckinsel needs to vote
this may need to be modified after James' merge... @tetron will take point on the updates

openapi/workflow_execution_service.swagger.yaml
```diff
@@ -495,10 +540,10 @@ definitions:
...
         type: string
         description: |-
-          OPTIONAL
-          The workflow CWL or WDL document, must provide either this or workflow_descriptor. When a base64 encoded gzip of
-          workflow descriptor files is offered, the `workflow_url` should be set to the relative path
-          of the main workflow descriptor.
+          REQUIRED
+          The workflow CWL or WDL document.
+          When workflow attachments files are provided, the `workflow_url` may be a relative path
+          corresponding to one of the attachments.
```
mckinsel: Would it be clearer to just say that `workflow_url` should match one of the filename parameters in the Content-Disposition headers?

openapi/workflow_execution_service.swagger.yaml
```diff
@@ -142,12 +166,42 @@ paths:
...
+
+        - in: formData
+          name: workflow_type
+          type: string
+
+        - in: formData
+          name: workflow_type_version
+          type: string
+
+        - in: formData
```
david4096: Isn't it possible to send in a request body with multipart upload items? We might leave the WorkflowRunRequest untouched and offer the file upload as an option, as opposed to requiring sending strings to multipart file uploads.

geoffjentry: If I understand you correctly we already discussed that as an option and voted on this path

geoffjentry:
In particular directed at @tetron @jaeddy @mckinsel but also in general:

Am I right in thinking that this PR has an impact on `WorkflowRunRequest` in the `WorkflowLog` section? If so, should we tackle that here or after merging this? Considering how long it's taken to get consensus on this (I think it is mergeable now?) 

I think I'd prefer to clean up the `WorkflowLog` stuff separately - I have a feeling that there are other lurking ripple effects from this change and we can handle them all at once. For instance I came across this when updating the Cromwell WES code and noticing something didn't map cleanly anymore.

tetron:
@geoffjentry the impact is that it ends up being defined separately (redundantly) in the API doc.

tetron:
I'm going to merge this, we can always continue to refine it in another PR.

geoffjentry:
@tetron Definitely, wasn't suggesting it was tricky just that there are now other inaccuracies in the spec. Since you merged it already it looks like we're in agreement that it can be sorted out later :)

kislyuk:
Sorry I'm late to this discussion. Just wanted to throw the following out there: I think all modern APIs use JSON as the data interchange format, and the tooling for JSON-based I/O is continuously improving and being invested in. I think making the WES API spec require the use of multipart form data will negatively impact the adoption of the API.

(For reference, to see the I/O preferences we made in the HCA DSS API, here is its Swagger spec: https://github.com/HumanCellAtlas/data-store/blob/master/dss-api.yml)

tetron:
@kislyuk Is the HCA DSS API handling actual file content?  It looks to me like it is JSON objects containing URLs to files uploaded separately.  WES provides for uploading file content directly.  Also I noticed in that document you linked, it uses multipart framing for streaming notifications, not JSON.

kislyuk:
@tetron yes, it handles actual file content, but only by way of redirects. For webhook notifications, the preferred payload format is JSON, but one of our contributors decided to add multipart form POST support for legacy applications.

kislyuk:
In general, handling file content in-band in the API might be a design issue in and of itself. If a text file needs to be included though, it can always be included as a JSON string value.

PR #45, opened 2018-06-18
Add `contact_info` to ServiceInfo.

tetron:
This is simply an additional field to ServiceInfo that should provide a way to contact the operator of a given WES endpoint.

### DIFF
--- a/openapi/workflow_execution_service.swagger.yaml
+++ b/openapi/workflow_execution_service.swagger.yaml
@@ -353,7 +353,9 @@ definitions:
         type: string
         description: |-
           An email address or web page URL with contact information
-          for the operator of a specific WES endpoint.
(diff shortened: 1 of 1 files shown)


geoffjentry:
Did this come out of the security group's requirements?

mckinsel:
I think this was to answer "What mechanism will you implement to enable users of this product to report discovered vulnerabilities?"

tetron:
Yes.

geoffjentry:
Ok, that's what I figured. In that case I'd suggest being more explicit about the purpose. If I hadn't been on the call today I'd have assumed this is one of those rando things which is a good enough idea on the surface but where one can't do everything :)

tetron:
@geoffjentry I added another sentence, what do you think?

geoffjentry:
Looks good, thanks

PR #80, opened 2018-07-24
Remove RunDescription. Closes #79

### LINKED ISSUE #79: RunDescription and RunStatus are identical - should they be collapsed?
Curious if anyone had thoughts on this.

For practical purposes `RunStatus` and `RunDescription` appear to be the same. Does anyone oppose collapsing them into a single entity?

@tetron @mckinsel @jaeddy ?

geoffjentry:


openapi/workflow_execution_service.swagger.yaml
```diff
@@ -549,10 +539,11 @@ definitions:
     properties:
       run_id:
         type: string
-        title: workflow run ID
+        title: REQUIRED
```
david4096: I think you meant to use `required: true`. The title field changes the display of items and how they're accessed by downstream clients.

openapi/workflow_execution_service.swagger.yaml
```diff
@@ -549,10 +539,11 @@ definitions:
...
       state:
         $ref: '#/definitions/State'
-        title: state
+        title: REQUIRED
+    title: 'Small description of a workflow run, returned by server during listing'
```
david4096: I think this should be "description".

openapi/workflow_execution_service.swagger.yaml
geoffjentry: I was going to say that i was just copying what was there, but the new issues explain why that's bad :) Will do.

geoffjentry: Hmmm, is `required` legal here? Swagger keeps complaining, but I haven't found something definitive in the spec either way

jaeddy: I don't think `required: true` is valid for properties anymore. My understanding is that the spec follows JSON Schema:

```
properties:
  required:
    -run_id
  run_id:
    type: string
    description: workflow run ID
```

That is, an array at the top level of `properties`, rather than a field within each specific property definition.

I don't know when this change took effect, but I've seen it in most examples for OpenAPI 2.0 and 3.0.

jaeddy: Whoops, I lied. `required` should be on the *same* level as `properties` for object definitions/schemas:

```yaml
type: object
required:
  -run_id
properties:
  run_id:
    type: string
    description: workflow run ID
```

david4096: @jaeddy thanks for that! You are correct, two different ways depending on if you're in parameters or a JSON schema.

https://github.com/ga4gh/data-object-service-schemas/blob/master/openapi/data_object_service.swagger.yaml#L530

https://github.com/ga4gh/data-object-service-schemas/blob/master/openapi/data_object_service.swagger.yaml#L551

geoffjentry:
Ah, thanks @jaeddy and @david4096 ! How's this?

david4096:
Thanks!

geoffjentry:
@david4096 I don't think it's an issue here as it seemed straightforward but I think we should be getting the approval of the various thumb-givers (and non-thumb-downers) prior to merging things. IOW myself, @jaeddy, @mckinsel , @tetron and @briandoconnor / @dglazer for the other group. For instance we hadn't heard from @tetron at all.

tetron:
No objection to this PR in particular, but I agree we should be trying to follow the agreed on process.  Github has settings to require certain approvals before a PR can be merged, who has permissions to edit the settings, @rishidev ?

dglazer:
We could enforce it programatically, or we could just remind anyone with
merge privs to follow the process. Given how few merges (I think) are left
before 1.0, and that the process may change after that, I lean towards the
lower-tech answer. I'm open though.

geoffjentry:
@dglazer agree - it's low enough volume and low enough velocity that I think if something really **was** controversial would could easily back it out

PR #85, opened 2018-08-20
Swagger autodoc

jaeddy:
Uses the swagger2markup gradle plugin and asciidoctor to (1) automatically generates human-readable asciidoc files from the contract-first OpenAPI (swagger) yaml spec; and (2) incorporate
manual content to build an overall document in HTML and PDF formats. Docs build added to Travis* config, with compiled HTML pushed to gh-pages.

*Travis routine probably needs a bit of cleanup.


jaeddy:
@david4096 — specifically looking for some feedback/assistance on the TravisCI config. I'd like to run the gradle plugin to autogenerate docs from the swagger yaml, then compile everything into HTML and PDF files. After the build, I'm deploying the HTML to gh-pages and (nominally) pushing the built docs back to the repo. The deploy bit is currently using my GH access token, which probably isn't ideal.. And I'm not totally sure how to manage the latter step of pushing to the repo.

Oh, and at some point I mistakenly removed your deploy script for pypi... so need to add that back in. :)

docs/README.md
```diff
@@ -0,0 +1 @@
+View the full [Reference Documentation](https://jaeddy.github.io/workflow-execution-service-schemas/docs/) for the Workflow Execution Service API.
```
briandoconnor: Hey, shouldn't this be this repo and not yours? @jaeddy

dglazer: @jaeddy -- shouldn't this point to somewhere in the GA4GH repo, not to your separate development area?

jaeddy: That was vestigial from testing/demo-ing for the Monday call. Thanks for catching!

jaeddy: Yep! Fixed. Thanks!

jaeddy:
It was intentional to keep the generated .adoc, but I like your idea of making them transient better. The `front_matter.adoc` and `index.adoc` are still needed to add manual content and specify the order of everything.

dglazer:
Thanks @jaeddy, and agree that the two .adoc input files need be under source control. Smaller thoughts that we can address after this is merged:
- do we want index.{html,pdf} to be under source control? I lean towards no for the same reason, but don't feel very strongly
- do we have/want some way to auto-timestamp the generated pages with when they were built, so that someone who just clicks a link in the README can get a sense of whether they're current?

And of course I expect a few rounds of content cleanup once we see the final output, but that's easy.

Issue #96, opened 2018-09-11
Address PRC Concerns: WesObject 


rishidev:
The PRC highlighted requested that the use of WesObject be clarified. Section 7.11 of the PRC comments document https://docs.google.com/document/d/1ul9Q_k5mwgzE1orrBsfb1mXrYyVWLDakGBeNyEmZQK0/edit

states:
WesObject defined as “An arbitrary structured object.” Describe the difference between WesObject vs. string (application/json). Unify use of WesObject vs JSON input/output.
- Used for /runs outputs and for /run workflow_params, prefer fit-for-purpose Schemas. If nothing special, consider using string (application/json). 

At the meeting it was suggested the spec maintainers (@mckinsel, @geoffjentry, @tetron, @denis-yuen) can clarify this - if the WesObject is meant to be arbitrary string the change to string (application/json) should be made in a PR. If there are reasons for keeping WesObject they should be stated so that the reasoning can be communicated with the PRC.

 The implementations do not need to be adjusted to account for this pre-Basel meeting.

geoffjentry:
I mentioned on the call yesterday that I thought `WesObject` was actually more of an artifact of history than seeds for something to come and that we should just get rid of it. In particular, my recollection is that we were viewing it as "this **should** be a JSON object, but giving ourselves wiggle room" type of thing.

I did some digging and do think that's true. 

It showed up in [this PR](https://github.com/ga4gh/workflow-execution-service-schemas/commit/c8e3980cc33542877ad062bf6f77bb7400237db1#diff-b1f8a5e75ad10e8057f21f2be2c37f66L342) and a) is an artifact of the old protobuf structure and b) definitely pays lip service to the JSON object thought.

I'd recommend either of the following, on a case by case basis:

- Making them strings
- Defining them as JSON objects

rishidev:
PRC happy with the modifications on this

PR #99, opened 2018-09-17
removed WesObject from the schema

delagoya:
Resolves issue #96 by removing WesObject from the schema.

### DIFF
--- a/openapi/workflow_execution_service.swagger.yaml
+++ b/openapi/workflow_execution_service.swagger.yaml
@@ -485,17 +485,16 @@ definitions:
           $ref: '#/definitions/Log'
         title: 'the logs, and other key info like timing and exit code, for each step in the workflow run'
       outputs:
-        $ref: '#/definitions/WesObject'
-        title: the outputs
+        type: object
+        title: a JSON dict of the outputs
   RunRequest:
     type: object
     properties:
       workflow_params:
-        $ref: '#/definitions/WesObject'
+        type: object
         description: |-
           REQUIRED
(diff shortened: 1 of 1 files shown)


dglazer:
Thanks Angel. I'm pretty new to OpenAPI, but I would have expected the two references to WesObject to be replaced by `type: string` (not `type: object`). Is there a reason to use an object without properties, instead of a string?

delagoya:
It was my reading of the conversation during the call yesterday that consensus landed on a JSON object in lieu of string type. The rational was that a String would require inspection and proper deserialization on the consumer, whereas a JSON object is understood to be a single parser. 

Having said that, I am also new to Swagger/OpenAPI! In my naive understanding I was assuming that both of these would be valid parameter values: 

```json 
{
    "param1" : "value1", 
    "param2": 2
}
```
```json
{
    "param1" :  { 
         "sub_param1": "value1", 
         "sub_param2": "value2"
    }, 
    "param2":  [1, 2, 3]
}
```
 

So if this is not the right way of encoding a arbitrary JSON object as a parameter type, someone with actual experience should comment.

dglazer:
Poking around a bit more I think you're right. Be great to have some expert confirmation.

tetron:
> It was my reading of the conversation during the call yesterday that consensus landed on a JSON object in lieu of string type. The rational was that a String would require inspection and proper deserialization on the consumer, whereas a JSON object is understood to be a single parser.

That's correct.

> So if this is not the right way of encoding a arbitrary JSON object as a parameter type, someone with actual experience should comment.

Based on https://swagger.io/docs/specification/data-models/data-types/#free-form I think you have it right.

PR #101, opened 2018-09-22
Update workflow_execution_service.swagger.yaml

dglazer:
Work in progress to clarify documentation in response to security team questions.


briandoconnor:
I talked with David and Rishi today.  We need to:
* add statement to docs saying HTTPS must to be used for real system
* delete http as an option from the OpenAPI yaml

Issue #103, opened 2018-09-30
Add CANCELLING state


rishidev:
Still to be discussed was the question on Section 7.10. State
- Should we add a "CANCELING" state to reflect the fact that the request to cancel has been received but is still being processed?

briandoconnor:
See pull request https://github.com/ga4gh/workflow-execution-service-schemas/pull/104

jaeddy:
Haven't confirmed that any implementations support this, but it's in the spec.

geoffjentry:
@jaeddy thanks for poking at this, just yesterday I was lamenting that there was no `CANCELLING`, perhaps I should have taken a look at the contemporary spec :)

jaeddy:
Just make sure you spell it `CANCELING` ;-)

geoffjentry:
oh good gravy ....

PR #104, opened 2018-09-30
addressing PRC feedback

briandoconnor:
See https://github.com/ga4gh/workflow-execution-service-schemas/issues/103

### DIFF
--- a/openapi/workflow_execution_service.swagger.yaml
+++ b/openapi/workflow_execution_service.swagger.yaml
@@ -456,6 +456,7 @@ definitions:
       - EXECUTOR_ERROR
       - SYSTEM_ERROR
       - CANCELED
+      - CANCELLING
     default: UNKNOWN
     description: |-
       - UNKNOWN: The state of the task is unknown.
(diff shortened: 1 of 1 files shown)


openapi/workflow_execution_service.swagger.yaml
```diff
@@ -456,6 +456,7 @@ definitions:
       - EXECUTOR_ERROR
       - SYSTEM_ERROR
       - CANCELED
+      - CANCELLING
```
dglazer: We should use a single L in "CANCELING" to match the existing "CANCELED". (Per a quick web search, both words can be spelled either way, with a single L preferred in American English and a double L in British English. So let's be consistent.)

openapi/workflow_execution_service.swagger.yaml
```diff
@@ -456,6 +456,7 @@ definitions:
       - EXECUTOR_ERROR
       - SYSTEM_ERROR
       - CANCELED
+      - CANCELING
```
dglazer: I changed it to a single L to match "CANCELED". (Both spellings are okay for both words, so let's be consistent.)

PR #160, opened 2020-10-01
Add `execution_location` to calls

### LINKED ISSUE #158: Add concept of project or workspace to WES
The current published WES specification has no provision for organizing executions. Executions go into a flat space. A user may wish to organize the executions however. SB powered platforms have concepts of Projects, Terra has a concept of workspaces, for example. On HPC executions users may choose to persist execution products or logs or other meta information in a particular folder, or space.

The SB WES implementation currently uses a custom parameter `project` in the `workflow_params` field. It would be useful to standardize how to do this by having an optional parameter that can be used to instruct the engine where executions can be organized.

For example, we could have a parameter `execution_location` with type string which refers to a path that the host platform/executor will use to organize the executions.

This parameter should be present in all calls that 
1. Start executions
2. List executions

Thanks!

kaushik-work:
Closes #158

### DIFF
--- a/openapi/workflow_execution_service.openapi.yaml
+++ b/openapi/workflow_execution_service.openapi.yaml
@@ -117,6 +117,12 @@ paths:
         explode: true
         schema:
           type: string
+      - name: execution_location
+        in: query
+        description: OPTIONAL Platform specific path to a project/workfspace/folder where execution meta-formation is persisted.
+        schema:
+          type: string
+      
       responses:
         200:
           description: ''
@@ -183,6 +189,9 @@ paths:
             schema:
               type: object
               properties:
+                execution_location:
+                  description: OPTIONAL Platform specific path to a project/workfspace/folder where execution meta-formation is persisted.
+                  type: string
                 workflow_params:
                   type: string
                 workflow_type:


ruchim:
hey @kaushik-work -- it sounds like we're introducing a concept of "working directory" into the WES implementation -- which is a utility that lets a client dictate "where" the outputs of a run go. Does this imply anything other than a directory on a local filestore or a pseudo-directory type thing in a cloud storage system? 

There might an access notion we have to solve for here as well -- for example is the WES implementation using the same token to do this move, or if that token doesn't have the scopes, how do additional tokens get added?

ruchim:
@ianfore you may have opinions on this -- a cross-section of cloud buckets x workflow outputs

ruchim:
@kellrott do you think this will be relevant to TES?

patmagee:
@ruchim this seems less like the idea of "put workflow outputs here" and more like the idea of "execute in this sub directory of a multi-tenant system". Ie, a WES implementor *WESICORP* has a multi-tenant system where users have projects (think like github). When a user would go to interact with the workflow they would need to supply a new parameter, to identify the "project" that they are interacting with, ie:

```http
POST https://wesicorp.com/ga4gh/wes/v1/runs .... execution_location=MY-PROJECT
GET https://wesicorp.com/ga4gh/wes/v1/runs?execution_location=MY-PROJECT
GET https://wesicorp.com/ga4gh/wes/v1/runs/8191?execution_location=MY-PROJECT
GET https://wesicorp.com/ga4gh/wes/v1/runs/8191/status?execution_location=MY-PROJECT
POST https://wesicorp.com/ga4gh/wes/v1/runs/8191/cancel?execution_location=MY-PROJECT
```

@kaushik-work did I capture this correctly? To me this feels like the spec would be sacrificing a bit of usability to serve the multi-tenant system, which may only be an issue for a very few number of engines. In fact this is achievable with the current WES schema, just in a different way. 

Another solution which would not require any spec changes at all would be loosen the requirement for `wes` to be at the root of a url. For example a multi-tenant system can easily serve user or project specific requests by embedding a WES api at a different Base URI:

```http
POST https://wesicorp.com/projects/8191/ga4gh/wes/v1/runs .... execution_location=MY-PROJECT
GET https://wesicorp.com/projects/8191/ga4gh/wes/v1/runs?execution_location=MY-PROJECT
GET https://wesicorp.com/projects/8191/ga4gh/wes/v1/runs/8191?execution_location=MY-PROJECT
GET https://wesicorp.com/projects/8191/ga4gh/wes/v1/runs/8191/status?execution_location=MY-PROJECT
POST https://wesicorp.com/projects/8191/ga4gh/wes/v1/runs/8191/cancel?execution_location=MY-PROJECT
```
Consumers of this API then just set the `WES_BASE_URI` to `https://wesicorp.com/projects/8191` instead of needing to specify the query parameter with each request. For people who argue about discoverability, we could leverage a service registry here at `https://wesicorp.com/services` which publishes all of the WES apis a user has access to

kellrott:
Regarding the question about how connected this is to TES, I think this would inform the workflow engine, which then create input/output paths for TES based on that configuration. All TES would need is the destination of the outputs and the credentials to write there. So I don't think this concept would need to be reflected into TES.

kghose:
@patmagee this is exactly right. I would envision this to be an optional parameter, so an addition to the spec. However, I will not be going on with this project. I will tag team @mattions in for when he is back from leave.

ruchim:
@patmagee 

So in this expression``execution_location=MY-PROJECT` 
On the left hand side -- the name "execution_location" implies a location somewhere in the cloud (be it bucket or some made up directory concept) 
On the right hand side, the word "my-project" indicates like an alias of style "project name". This is more like a string name for something -- and possibly the engine knows to convert that string into a cloud location. Using your github anology of repositories -- is the idea that my-project is a reference to an entity somewhere on the internet? or is it just a string? 

If the My-Project is a string/tag -- then the key shouldn't be "execution_location" as there is no location. If the My-project is really the location of a logical resource -- like "my-bucket-name" -- then that makes more sense. I feel one wouldn't want to expose names of resources in an API like this -- as that wouldn't be secure and sort of expose the data you may be trying to keep private.

patmagee:
@ruchim I think this is less analogous to a bucket and more a gcp project (or workspace for the cromwell use case), or azure subscription, or seven bridges project.  Another term I think that would be more explicit would  instead be `realm`, or `workspace` or `project`.

As @kghose is suggesting in this pr, every api request would need the query parameter set to define the location. Imo that's not great api design for the user and I'm not really in favor of it

Ultimately I think this would only map onto a large multi tenant system, and even there it's possible their implementationa are different enough that this does not really capture their use.

ruchim:
Maybe worth putting this PR on pause -- until we know who the new owner is.

patmagee:
closing this since the original repo no longer exists

Issue #159, opened 2020-10-01
Use more standard paging for `GET /runs` endpoint


kaushik-work:
The `GET /runs` endpoint (https://ga4gh.github.io/workflow-execution-service-schemas/docs/#operation/ListRuns) uses an unorthodox pagination method.

It would be better to use more standard pagination, such as that suggested [here](https://jsonapi.org/format/#fetching-pagination)

ruchim:
thanks @kaushik-work -- will check with the TES team to make sure we have consistency in pagination across TES & WES.

ruchim:
@kellrott -- is this an example of where WES could learn from what the TES standard is doing? Or is this a problem that needs solving for TES as well?

patmagee:
FWIW Search Has implemented somthing similar to this [here](https://github.com/ga4gh-discovery/ga4gh-search/blob/develop/openapi/openapi.yaml#L269)

```json
{
  "pagination": {
    "next_page_url": "http://.....",
    "previous_page_url": "http://...."
  }
}
```

denis-yuen:
TRS uses HTTP headers for pagination links, similar to GitHub's API 
https://github.com/ga4gh/tool-registry-service-schemas/blob/2.0.0/openapi/openapi.yaml#L209
https://docs.github.com/en/free-pro-team@latest/rest/guides/traversing-with-pagination

Mifrill:
any updates on that?

As an alternative it also might be useful to add `total_pages` or `total_runs` as new property for endpoint response:
https://github.com/ga4gh/workflow-execution-service-schemas/blob/c3b19854240c4fcbaf3483e22b19db0a918a7ee5/openapi/components/schemas/RunListResponse.yaml#L3-L11
or like it's done in TRS via header:
```yml
last_page:
  description: A URL that can be used to reach the last page based on the current
    page record limit.
  schema:
    type: string
```
https://github.com/ga4gh/tool-registry-service-schemas/blob/2ecf30ebd877256d521c5d82db577c69decf0b1b/openapi/openapi.yaml#L181-L185

since currently there is no way to implement "orthodox" pagiantion with `first / next / last` references for pages, only possible to implement Next or Prev page, that might be solved by addition total number of runs.

Mifrill:
@denis-yuen 
> TRS uses HTTP headers for pagination links

That's correct, but why WES is not?
[...]

PR #183, opened 2022-08-01
feat: add total_runs in response of GET /runs endpoint

Mifrill:
The property `total_runs` allows to implement the pagination with the references for each page, includes the last page and showing number of pages.

Related to https://github.com/ga4gh/workflow-execution-service-schemas/issues/159#issuecomment-1184525759


patmagee:
@Mifrill thanks for the PR! I would be really interested to know your use case and what you are trying to solve with the `total_runs` (outside of the conversation on the linked issue). I do not think this would add TOO much of a burden on engines, although I will say in some cases it is simply not possible to know the total number of runs without performing an extremely expensive operation. So I wonder if this is a required field, or if maybe there is a generic way to represent optional summary metadata in the `RunListResponse`.

elasticsearch has the concept of `"hits"` which may not really be applicable here since this is not a search context, but maybe `"metadata"` or some other field:

```json
{
   "runs": [],
   "metadata": {
     "total_runs": 5000
   }
}

**Comment by cjllanwarne** [comment]
+1 this seems pretty important for neat pagination

File: openapi/components/schemas/RunListResponse.yaml [hunk 1]
```diff
@@ -9,4 +9,8 @@ properties:
   next_page_token:
     type: string
     description: A token which may be supplied as `page_token` in workflow run list request to get the next page of results.  An empty string indicates there are no more items to return.
+  total_runs:
+    type: integer
+    description: >-
+      Total count of workflow runs that the service has executed or is executing and the caller has permission to see.
```
cjllanwarne, line 15: Suggestion for brevity, and future-proofing (in case we ever add the ability to filter results from this endpoint). I'd also suggest replacing the words "workflow runs" with just "runs", per the WES terminology.

```suggestion
      Total count of runs in the result set.
```

Mifrill:
@patmagee thanks for the review!

>  what you are trying to solve with the total_runs

The total count required to implement "orthodox" pagination, by dividing the total count by page size we can provide a direct link for any page. Rather than in the current implementation only possible to go next page or the previous page.

>  it is simply not possible to know the total number of runs without performing an extremely expensive operation

In my understanding, the implementation of the counter is the responsibility of each particular service, hence the possible performance issues should be handled on the specific service side. There are many possibilities to avoid performing expensive operations like `counter field` implementation or last-document-index etc. it fully depends on a particular service's inner logic, so it's not related to the schema.

> if maybe there is a generic way to represent optional summary metadata in the RunListResponse.

I don't think it is a good way to go because if we will follow that logic, the filed `next_page_token` also should be moved to some metadata wrapper field:
https://github.com/Mifrill/workflow-execution-service-schemas/blob/ea050992fd0d95a9624024bf2f14977a905ff9c7/openapi/components/schemas/RunListResponse.yaml#L9

openapi/components/schemas/RunListResponse.yaml
Mifrill: > in case we ever add the ability to filter results from this endpoint

not sure I follow.
The sentence `has executed or is executing` is looks generic in terms of `workflow-execution-service-schemas`.
The sentence `the caller has permission to see` didn't set any constraints in case of policy scope implementation, the records might be across entire data as well.

> "workflow runs" with just "runs", per the WES terminology

As I see in other examples, the `workflow runs` preferred:
Selection_11021
https://github.com/ga4gh/workflow-execution-service-schemas/blob/develop/openapi/components/schemas/RunListResponse.yaml


Selection_11022
https://github.com/ga4gh/workflow-execution-service-schemas/blob/develop/openapi/components/schemas/RunStatus.yaml

patmagee:
@Mifrill thanks for the clarification. I think your suggestion makes a lot of sense, and I wonder if there is a way to frame this PR in a forward looking way and consider what we would need to improve pagination in general (starting with a total_runs). There is [this open issue](https://github.com/ga4gh/workflow-execution-service-schemas/issues/159), but it has not really gotten any steam lately.  I know  trs (@denis-yuen) has a specific approach that the settled on which is externally documented, so I wonder if we should likewise investigate what paging schemas are there in the wild and what fits with WES in a holistic way

patmagee:
@Mifrill would you mind updating your changes against the current develop branch? We merged all of the openapi components into a single file to better work with our tooling

Mifrill:
@patmagee thanks for the ping 👍. Updated

uniqueg:
On this subject, I would like to draw attention to the [draft](https://docs.google.com/document/d/12Y0MgIRqaBt2CqzEscArVEkCS_8WHFRdQef9IFmEGVc/edit) of the upcoming GA4GH pagination style guide (being drafted to address https://github.com/ga4gh/TASC/issues/29).

The style guide is not very prescriptive at the moment (I think it's rather a survey of currently used pagination solutions across the GA4GH API landscape with a discussion of pros and cons) and is definitely open for feedback.

jaeddy:
FYI @uniqueg - I believe **[this](https://docs.google.com/document/d/1Zu_ksx6lG3OXcTTViIWyrSMqGTkcpDk9Oa7g1c7StvM/edit?usp=sharing)** is the latest version of the pagination guide from @andrewyatz and @mamanambiya.

andrewyatz:
> FYI @uniqueg - I believe **[this](https://docs.google.com/document/d/1Zu_ksx6lG3OXcTTViIWyrSMqGTkcpDk9Oa7g1c7StvM/edit?usp=sharing)** is the latest version of the pagination guide from @andrewyatz and @mamanambiya.

Yes this is the right one

uniqueg:
Thanks for pointing that out @jaeddy 🙏

Mifrill:
Hey @patmagee, @cjllanwarne just a gentle reminder, please have a look at this PR when you get a time 🙏

Issue #134, opened 2019-05-09
Frozen versions of WES spec


susheel:
To backward compatibility support WES servers/clients, we need to provide a frozen-in-time version of the spec. Suggest we keep a major-minor-version in the `openapi` folder for the repo locked for any further edits.

`WES v1` --> `/openapi/`**`v1`**`/workflow_execution_service.swagger.yaml`
`WES latest` --> `/openapi/workflow_execution_service.swagger.yaml`

jaeddy:
I like the first option. When you say major-minor, do you mean there might be a `/openapi/v1.1/workflow_execution_service.swagger.yaml` as well?

I have no objections to dropping `swagger` from the filename. Is there any convention or standard we should follow?

susheel:
I don't think there is a convention. I was suggesting v1 and v1.1, but happy to limit to just major releases.

geoffjentry:
IMO the general convention is `vMAJOR`, not `vMAJOR.MINOR` in the URL

susheel:
Happy with `vMAJOR`

jaeddy:
Sounds good. @susheel, can you make a PR for this?

juhtornr:
I think that we should fix #139 before freezing the spec.

ruchim:
@susheel -- sorry to get to this so late -- does https://github.com/ga4gh/workflow-execution-service-schemas/blob/develop/openapi/workflow_execution_service.openapi.yaml#L5 solve the spirit of this request -- that we track major/minor version along with the exact definition of that spec?

vinjana:
The missing 1.0.1 version tag issue is clearer covered in #191. 
Semantic versioning is also covered in #71.

I don't see what other topics of this issue are not better covered by other issues, and propose to close this issue.

uniqueg:
Thanks for pointing me to your comment here, @vinjana, I hadn't seen that.

As for closing this issue: I think the proposal to include frozen/tagged/released versions of the specs in separate files goes beyond what is being described in #191. Having all versions available in the same directory might be more convenient in some cases (e.g., when comparing versions) as opposed to having to check out releases individually. However, it also comes with additional complexity, on the maintenance side, but potentially also on the user/implementer side (we probably wouldn't want new implementers to think much about what version to use, but instead rather pick the latest one).

[...]

PR #189, opened 2022-09-23
Remove Definition and Reference Python Code

patmagee:
The swagger definition of WES has been deprecated in favor of openapi 3.0, however it was kept in the repo for backwards compatibility. Speaking with a number of people at plenary it came up that instead of being helpful, this has actually caused some issues among implementors. Since the swagger API is outdated it can easily lead to implementing the "WRONG" api. Additionally (I have felt this as well), there is no clear direction of what spec doc should be modified when implementing a PR leading to implementors editing both docs.

Additionally, I have removed the Reference python code. This code used to be built as part of the travis build process, but has been commented out for sometime and is outdated with the current spec. This has also been superseded by the WES GA4GH starter kit (https://github.com/ga4gh/ga4gh-starter-kit-wes), which is more actively maintained and updated.


jaeddy (approved):
I think preserving the Swagger spec was related to [this issue](https://github.com/ga4gh/workflow-execution-service-schemas/issues/134), but I agree that it makes more sense to only keep the OpenAPI from the current version moving forward.

If anyone ever wants to look at the original Swagger from the v1.0 release, it can be found on the tag branch or in Zenodo.

wleepang (approved):

patmagee:
@jaeddy @wleepang I apologize, it looks like local state from a divergent develop branch sneaked its way in. I have updated the PR to remove those extra files, if you to take one additional view

wleepang (approved):

jaeddy (approved):
Was wondering about the Workflow Inputs stuff — makes more sense now. 🙂

uniqueg:
Yes - finally!! 😆

PR #196, opened 2022-10-21
Fix/openapi definition

patmagee:
This PR Aims to accomplish two things:

1. Fix the OpenAPI validation errors that are now being thrown from our newly setup CI/CD pipeline (Thank you @vsmalladi)
2. Drastically reduce the complexity of the OpenAPI documentation directory and further reduce redundancy (thus reducing errors). I anticipate open PR's will likely need to be updated to work with the new setup.  I think the original idea of componentizing the spec was great, however for an API surface as small as WES, it introduced unnecessary complexity and confusion. Having a single file that contributors need to edit leaves no confusion as to what they are responsible for doing in order to update the spec. To that end, I additionally removed the `json` version of the openapi spec. While I can understand the motivation to have  this file as well, there are enough tools to convert between yaml and json that I think it is also redundant.


patmagee:
@wleepang after some debugging with @vsmalladi it appears that develop is actually broken and will now work with the GitHub actions.

This current branch does seem to work (we've tested making a pr to it). I propose we merge this change bypassing the checks this once

wleepang (approved):
Two way door to merge and adjust CI testing as this is not a release

Issue #191, opened 2022-09-27
No tag/release for version 1.0.1


uniqueg:
The [reference documentation](https://ga4gh.github.io/workflow-execution-service-schemas/docs/) linked to in this repository's `README.md` is for version `1.0.1` of the specification. However, there is no corresponding [release](https://github.com/ga4gh/workflow-execution-service-schemas/releases) and/or [tag](https://github.com/ga4gh/workflow-execution-service-schemas/tags) for version `1.0.1` (latest tag and release are for `v1.0.0`).

Moreover, I seem to be unable to find the commit from which the specification linked to at the top of the reference documentation is derived. While the version string in the specification itself seems to have been bumped in [this commit](https://github.com/ga4gh/workflow-execution-service-schemas/commit/a9b74ebd81e70a69ea9f0bf1de64e32cf5abb6f0), the (state of the) specification in that commit does not seem to match the specification linked to in the reference documentation.

As it is of critical importance that implementers choose a definite version of a specification to develop against, it would be great to:
- Identify what exactly `v1.0.1` constitutes
- Tag the corresponding commit
- Create an associated release
- Ensure that the version of the specification linked to in the reference documentation matches the one in that release

Compare also #192 and #193

vinjana:
In [this comment](https://github.com/ga4gh/workflow-execution-service-schemas/issues/134#issuecomment-1258164027) I refer to the versioning scheme. I just want to mention it here because, I think, if we want semantic versioning (SemVer2) then the version that is now published in the reference documentation as 1.0.1 should actually be 1.1.0, because it adds fields to the `ServiceInfo`, and API additions qualify for minor bumps according to SemVer2. I think it would be fine to tag this version both 1.0.1 and 1.1.0 and document this in the changelogs. Just an idea.

uniqueg:
Thanks @vinjana. Agree on SemVer, and I think that was the original intention. Given that there is not a tag/release yet for `v1.0.1`, I think it's indeed an opportunity to change the version to `v1.1.0` and then create the release.

I also agree with you on the more general points on having a clear governance regarding the creation of releases (ideally across -at least- the GA4GH Cloud WS APIs).

vinjana:
Duplicate of #157 .

uniqueg:
Oops, sorry. But guess it's time to track down the relevant 1.0.1 commit then :)

vinjana:
[...]

PR #200, opened 2023-01-08
merge master back into develop

wleepang:
The last commit on `master` is the commit that matches the nominal `1.0.1` release and the current specification docs per #191 . This pull request merges that release back into `develop` as the two branches have diverged since then.


uniqueg:
Seems like only the commit, but not the content of the commit is missing, right? In that case, I think it should be noncontroversial to merge

patmagee (approved):

uniqueg:
Fantastic! FYI, @patmagee and @wleepang: I have created a release for `1.0.1`: https://github.com/ga4gh/workflow-execution-service-schemas/releases/tag/1.0.1

I have used the autogenerate feature to create the release notes. Given that the issue/PR descriptions are mostly not super descriptive, those aren't very good, but I feel it's better than nothing. Maybe at some point we could clean up the release notes for the different releases, or at leas ttry to keep this a bit cleaner for the latest release (`1.1.0`) and moving forward. Personally, I would recommend that we use [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/) messages for commits to the `develop` branch. We achive this by squashing commits from PRs from feature branches and using conventional commit messages as PR titles.

PR #201, opened 2023-01-09
A simple change

patmagee:


uniqueg:
I guess your test failed, @patmagee 😆

patmagee:
@uniqueg i am open to suggestions! I am trying to get the actions working again, but there seems to be a permissions issue

uniqueg:
I think the problem is for PRs that come from forks. I don't know what the specific issue is, but I think we had the same in TES, and when opening a PR from a feature branch it went away. Of course accepting only Para from feature branches is not a sustainable solution, but the behavior may perhaps offer some clues as to what may be the problem.

PR #202, opened 2023-01-10
Changed actions to use pull_request_target

patmagee:


uniqueg:
So this fixed the issue? :)

patmagee:
Yes! But in order for it to have taken effect, it actually needs to be in the target branch, hence the rapid merge haha

PR #205, opened 2023-04-01
Feat: Make Tasks Addressable by ID and extend available information

### LINKED ISSUE #204: Add unique identifier to tasks
## Why

When polling workflow state it is often helpful to monitor the progress of a given task over time. Unfortunately there currently is no guarantee that any identifier present in the `task_logs` or list returned by the `/tasks` endpoint and you can never be sure that the task you are looking at is the same one from the preview retrieval

Additionally, Since merging: https://github.com/ga4gh/workflow-execution-service-schemas/pull/177 into the spec the goal is to make individual tasks addressable in the future. 

## Proposal

1. We can add a new `task_id` attribute within the `Log` object. I prefer this approach since it can be an easy transition to presenting `TES` information. If we do not want to return TES objects directly, we could also include a link `tes_url` to the task definition if we ever wanted to go that way.. but that is all for future consideration, adding an ID is minor bridge there
2. We can add verbage stating the `name` of the task should be Unique across all runs. IMO this is the inferior option, since the name often has some sort of meaning to the end user. Requiring this attribute to be unique now may require the engine to obscure the original name

patmagee:
## What
This PR is an implementation of the discussion in #204. 

- It adds a new model called `TaskLog` which is distinct from the `Log` model. The rational here is that having a specific model for task information allows us to put more task oriented data in it. Properties like `id` and `tes_uri` are strictly related to tasks, since `run_id` already is defined in the outer scope of a workflow. 
  - I have explicitly chosen to keep the type of `task_logs` within the `RunLog` as `Array[Log]` to prevent any breaking change.
  - This was motivated by the fact that an `id` MUST be required if we are try rely on it for addressing tasks. Requiring this within the `Log` attribute did not make sense since it was shared between the workflow and individual tasks. Additionally, forcing a required attribute could break downstream implementations
- Adds a new endpoint for retrieving a task directly. Tasks are addressable now so why not! In a system that has 100s or 1000s of tasks for a given workflow run, this subtle optimization could make life a lot easier
- added a new `tes_uri` field to allow for interop between tes and wes. This approach does not try to incorperate TES into the WES sep directly but provides an easy and Official mechanism to go between the two. TES is a much more rich view of a task so we should not try to re-implement
- add `system_logs` properties to `Log` and `TaskLog` entities. 

Closes #204


openapi/workflow_execution_service.openapi.yaml
```diff
@@ -788,14 +867,67 @@ components:
...
-            $ref: '#/components/schemas/Log'
+            $ref: '#/components/schemas/TaskLog'
           description: The logs, and other key info like timing and exit code, for each step in the workflow run.
         next_page_token:
           type: string
           description: >-
             A token which may be supplied as `page_token` in workflow run task list request to get the next page
             of results.  An empty string indicates there are no more items to return.
       description: The service will return a TaskListResponse when receiving a successful TaskListRequest.
+    TaskLog:
```
uniqueg, line 878: Since this seems to be strictly a superset of `Log` and is also _intended_ to be a superset of `Log` (at least until a hypothetical future in which this functionality would be provided by TES or externalized, generic WES/TES schemas; which would require breaking changes anyway), shouldn't we then rather make use of `allOf` to _extend_ the existing `Log` schema with the new properties and thus avoid repetition?

openapi/workflow_execution_service.openapi.yaml
```diff
@@ -788,14 +867,67 @@ components:
...
+      title: TaskLog
+      type: object
+      properties:
+        id:
+          type: string
+          description: A unique identifier which may be used to reference the task
```
uniqueg, line 884: I have a feeling that `id` should be apart from this schema, similar to `RunStatus`. It would make more sense to me for the new endpoint. If we already supply the idea to get the task info, why would we need it in the response? And if we get a list of tasks, it's also maybe clearer to return, for each task, its ID and its log. I feel that this is semantically more appropriate as well.

But maybe I'm missing something!

Incidentally: I have just seen that there is some really weird regression (at least I think it's regression, because I'm pretty sure it used to be different, but too lazy to check now) in `RunListResponse`. I think it did (and it should) return an array of `RunStatus` (which is a run identifier and a state, basically), _not_ `State`. What good is an array of just _states_, if they don't link to specific runs??

Anyway, apart from the fact that this appears to be degraded, I think that same pattern should be used for the task logs as well.

uniqueg (changes requested):

uniqueg:
I've just seen that I noticed the issue with `RunStatus` vs `State` before and that it should be fixed with #172 (I will look at that tomorrow). So please ignore that part. I guess it's just important that we merge #172 first and then merge that into this PR before merging.

openapi/workflow_execution_service.openapi.yaml
patmagee: Imo `TaskLog` is more akin to `RunRequest` then `Log` or `RunSummary`. In this case the `TaskLog` IS the entity according to REST semantics, and there is not really any other place to put the identifier.

In most REST systems, responses that are a resource, typically have all of the relevant information pertaining to that resource. This may include an ID, but it also may include things like a `self_link` or other identifiers. I think NOT having the ID here actualy reduces the fucntionality of this PR, since the whole ID is to give tasks a stable identifier across runs.  If the ID is not part of the listing operation, then where would it come from?

PR #206, opened 2023-04-02
fix: return RunStatus, not State, in RunListResponse

uniqueg:
Addresses the regression described [here](https://github.com/ga4gh/workflow-execution-service-schemas/pull/172#discussion_r1068109227) and [here](https://github.com/ga4gh/workflow-execution-service-schemas/pull/172#discussion_r1068147005).


patmagee (approved):
LGTM, I am happy to merge this if you are happy with it

PR #207, opened 2023-04-02
feat: add alternative schema for array items in RunListResponse

### LINKED ISSUE #165: Proposal: Add timestamps to RunStatus
Currently the `RunStatus` Object is quite minimal, comprised only of two keys: `state` and `run_id`. As a user, this provides me with very little context of the runs and makes creating UI's or CLI's awkward and hard to read. 

I propose we add two keys `start_time` and `end_time` to the `RunStatus`. This information is readily available already as part of the `RunLog` and should not pose a significant challenge for implementors to retrieve. While this does not provide complete context to the user it is a vast improvement over what we currently have and at least they will be able to see:

1. When a run was submitted
2. How long a run has been executing for
3. Provide an easy way of sorting runs by start or end time.

```json
{
  "run_id":"ff07e7a8-4da5-4e4f-b6a7-164cf4d6b45e",
  "status": "RUNNING",
  "start_time": "2021-01-07T00:00:000Z",
  "end_time":  "2021-01-07T07:21:000Z"
}
```

uniqueg:
Alternative to #172 
Closes #165

This PR differs from #172 in the following ways:
- A **separate schema `RunSummary`** is defined, to be used specifically for the `runs` array items in `RunListResponse`
- `RunSummary` **extends `RunStatus` (not modified!) with `required` additional fields**
- For backward compatibility, either **`RunSummary` or `RunStatus` (`anyOf`) are acceptable** responses for `GET /runs`
- Next to `start_time` and `end_time`, also workflow `name` and `tags` are included in the extended response; `workflow_url` was not included, as it can point to a local object and its support would therefore require WES to serve objects (up for debate of course, following #172 and #165)

Taken together, this proposal avoids or partly addresses the following [concerns raised against #172](https://github.com/ga4gh/workflow-execution-service-schemas/pull/172#pullrequestreview-1368081769):
- ***"Proposal has side effects:"*** By using a separate schema to accommodate the additional metadata fields to be returned in run list responses, the `RunStatus` schema remains untouched, thus retaining the behavior of `GET /runs/{run_id}/status` (which also uses `RunStatus`).
- ***"Implementers may choose to ignore added fields":*** While implementers can still return `RunStatus` instead of `RunSummary` (or indeed any other response that has the `run_id` and `state` properties) in run list responses for backward compatibility (`anyOf`), this PR adds a couple of measures to more strongly encourage the use of the new extended schema. In particular,
  - A `DEPRECIATION WARNING` is added to the description of `RunListResponse`, informing implementers that the use of `RunStatus` in `RunListResponse` will be discontinued in favor of `RunSchema` only.
  - The new schema ***requires*** the additional fields in `RunSummary`, which - together with the depreciation warning - should tell every implementer clearly where the specs will be going in this respect; indeed, as soon as support for `RunStatus` in `RunListResponse` is removed (a simple change), any compliant implementation MUST provide these additional fields in their run list responses.

**It does NOT address the concern that _"clients may not agree with selection of fields"_**.

### DIFF
--- a/openapi/workflow_execution_service.openapi.yaml
+++ b/openapi/workflow_execution_service.openapi.yaml
@@ -651,9 +651,6 @@ components:
         - $ref: '#/components/schemas/RunStatus'
         - type: object
           properties:
-            name:
-              type: string
-              description: The workflow name
             start_time:
               type: string
               description: When the run started executing, in ISO 8601 format "%Y-%m-%dT%H:%M:%SZ"
@@ -666,9 +663,6 @@ components:
               additionalProperties:
                 type: string
           required:
-            - name
-            - start_time
-            - end_time
             - tags
       description: Small description of a workflow run
     RunRequest:
--- a/openapi/workflow_execution_service.openapi.yaml
+++ b/openapi/workflow_execution_service.openapi.yaml
@@ -71,6 +71,10 @@ tags:
     x-displayName: RunStatus
     description: |
       <SchemaDefinition schemaRef="#/components/schemas/RunStatus" />
+  - name: runsummary_model
+    x-displayName: RunSummary
+    description: |
+      <SchemaDefinition schemaRef="#/components/schemas/RunSummary" />
   - name: runrequest_model
     x-displayName: RunRequest
     description: |
@@ -111,6 +115,7 @@ x-tagGroups:
       - runid_model
       - state_model
       - runstatus_model
+      - runsummary_model
       - runrequest_model
       - runlog_model
       - log_model
@@ -639,7 +644,33 @@ components:
           type: string
         state:
           $ref: '#/components/schemas/State'
-      description: Small description of a workflow run, returned by server during listing
+      description: State information of a workflow run
+    RunSummary:
+      title: RunSummary
+      allOf:
+        - $ref: '#/components/schemas/RunStatus'
+        - type: object
+          properties:
+            name:
+              type: string
+              description: The workflow name
+            start_time:
+              type: string
+              description: When the run started executing, in ISO 8601 format "%Y-%m-%dT%H:%M:%SZ"
+            end_time:
+              type: string
+              description: When the run stopped executing (completed, failed, or cancelled), in ISO 8601 format "%Y-%m-%dT%H:%M:%SZ"
+            tags:
+              type: object
+              description: Arbitrary key/value tags added by the client during run creation 
+              additionalProperties:
+                type: string
+          required:
+            - name
+            - start_time
+            - end_time
+            - tags
+      description: Small description of a workflow run
     RunRequest:
       title: RunRequest
       type: object
@@ -813,12 +844,18 @@ components:
         runs:
           type: array
           items:
-            $ref: '#/components/schemas/RunStatus'
+            anyOf:
+              - $ref: '#/components/schemas/RunStatus'
+              - $ref: '#/components/schemas/RunSummary'
           description: A list of workflow runs that the service has executed or is executing. The list is filtered to only include runs that the caller has permission to see.
         next_page_token:
           type: string
           description: A token which may be supplied as `page_token` in workflow run list request to get the next page of results.  An empty string indicates there are no more items to return.
-      description: The service will return a RunListResponse when receiving a successful RunListRequest.
+      description: >
+              The service will return a RunListResponse when receiving a successful RunListRequest.
+              DEPRECIATION WARNING: The use of `RunStatus` as the schema for `runs` array items will
+              not be permitted from the next major version of the specification (2.0.0) onwards. We
+              encourage implementers to use `RunSummary` instead.
     ErrorResponse:
       title: ErrorResponse
       type: object


openapi/workflow_execution_service.openapi.yaml
```diff
@@ -639,7 +644,33 @@ components:
...
+            name:
+              type: string
+              description: The workflow name
+            start_time:
+              type: string
+              description: When the run started executing, in ISO 8601 format "%Y-%m-%dT%H:%M:%SZ"
+            end_time:
+              type: string
+              description: When the run stopped executing (completed, failed, or cancelled), in ISO 8601 format "%Y-%m-%dT%H:%M:%SZ"
+            tags:
```
patmagee, line 663: I am in favour of this , but I would want to check with other implementors of WES (ie @wleepang) to see if this is a non-starter for them.

openapi/workflow_execution_service.openapi.yaml
```diff
@@ -639,7 +644,33 @@ components:
...
           $ref: '#/components/schemas/State'
-      description: Small description of a workflow run, returned by server during listing
+      description: State information of a workflow run
+    RunSummary:
+      title: RunSummary
+      allOf:
+        - $ref: '#/components/schemas/RunStatus'
+        - type: object
+          properties:
+            name:
```
patmagee, line 654: Similar to tags, I am largely in favour of this attribute, but I would probably want to ensure this is accessible by implementors before adding it

openapi/workflow_execution_service.openapi.yaml
```diff
@@ -813,12 +844,18 @@ components:
         runs:
           type: array
           items:
-            $ref: '#/components/schemas/State'
+            anyOf:
+              - $ref: '#/components/schemas/RunStatus'
+              - $ref: '#/components/schemas/RunSummary'
```
patmagee, line 849: I like this approach, since it does not break the API, but strongly suggests to implementors to use the new approach

patmagee (approved):
Thank you @uniqueg For putting this together! I think it has succinctly summarized the entirety of our discussion in the other thread in a very clean and concise way.

Overall I am in favour of this change! I addressed my original concerns and solves a few complexities with the breaking changes my PR introduced

uniqueg:
Thanks @patmagee :) I would still like to prepare a PR for the solution proposed [here](https://github.com/ga4gh/workflow-execution-service-schemas/pull/172#pullrequestreview-1368081769), before reaching out to implementers to vote. But I'm afraid it's gonna take me another two weeks or so.

patmagee:
@uniqueg can you please rebase this on develop

patmagee:
@wleepang @cjllanwarne I wonder what your thoughts are on this modified PR?

openapi/workflow_execution_service.openapi.yaml
```diff
@@ -639,7 +644,33 @@ components:
...
+            tags:
+              type: object
+              description: Arbitrary key/value tags added by the client during run creation 
+              additionalProperties:
+                type: string
+          required:
+            - name
+            - start_time
+            - end_time
+            - tags
```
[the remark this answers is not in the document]
uniqueg: You are probably right.

But then we could set defaults for these values (e.g., empty strings). That way we would perhaps encourage implementers a little more to give their best at producing these values, wherever possible. And on the client side we could rely on these fields being available.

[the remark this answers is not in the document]
patmagee: In general I prefer `null` to an empty value like `""` (empty maps are fine).  Empty strings end up requiring a bunch of edge case handling on the client side, ie in typed systems you now need to ensure that non null values can actually be deserialized as the expected type.

For example, I can represent a `RunSummary` in java like the following: 
```java
public record RunSummary(String id, String name, Instant start_time, Instant end_time, Map<String,String> tags){
}
```

Now, if I actually wanted to use this record when receiving responses, I would not be able to directly deserialize the value directly, but would now need to have a "holding" class because `""` is not a valid Date. Trying to deserialize the following using a library like `Jackson` would result in a `JsonProcessingException`, OR a `DateTimeParseException`. 
```
{
  "id": "foo",
  "name": "bar",
  "start_time": "",
  "end_time": ""
  "tags": {}
} 
```

Whereas the following would work:
```
{
  "id": "foo",
  "name": "bar",
  "start_time": null,
  "end_time": null
  "tags": {}
} 
```

openapi/workflow_execution_service.openapi.yaml
wleepang: Where would this come from? The current `RunRequest` doesn't have an explicit attribute for `name`. IIRC, the way we've handled it is with a specific `tag`. That said, if `name` is important enough to be used by other parts of the API, it might be worth elevating it to a top level property of `RunRequest`.

openapi/workflow_execution_service.openapi.yaml
wleepang: Agree

openapi/workflow_execution_service.openapi.yaml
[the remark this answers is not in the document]
wleepang: I don't think `name` requires parsing of a workflow. As I mentioned on a separate comment, `name` is effectively a property of `RunRequest` (possibly defined as a `tag`) - that is it is the name of the **run** and not the name of the **workflow**. For that I think I'll backtrack a bit and say that `tags` is the better place to hold this information and that **both** a run name and workflow name are a good idea.

For example:
- workflow names can be:
    - nf-core-sarek-3.1.2
    - gatk4-variant-discovery
- run names can be:
    - sample-id-12345
    - lims-id-{{uuid}}

Having these as tags presents a way to filter the runs list response along the lines of:
- "how many runs of a specific workflow are there?"
- "how many workflows or runs are processing data for a specific sample?"

openapi/workflow_execution_service.openapi.yaml
uniqueg: It's a property of the [`Log` model](https://github.com/ga4gh/workflow-execution-service-schemas/blob/6e6327fa680cfb03d8e81bf594b7625b5348e011/openapi/workflow_execution_service.openapi.yaml#L731-L759):

```yaml
    Log:
      title: Log
      type: object
      properties:
        name:
          type: string
          description: The task or workflow name
```

wleepang:
I think `tags` is more important than `name` for this response. (see my [comment](https://github.com/ga4gh/workflow-execution-service-schemas/pull/207#discussion_r1165734575) above).

From a client perspective, `runid`, `(start|end)_time`, and `status` are certainly must haves. I'd say `tags` is required and it is **strongly** recommended that "run_name" and "workflow_name" tags are included.

openapi/workflow_execution_service.openapi.yaml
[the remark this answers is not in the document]
uniqueg: We would already have unique identifiers for runs (the `run_id`). Having an optional human-readable alternative to these is useful, I think, but would indeed go well in `tags`, as it's really just an alternative.

Workflow names, on the other hand, are extremely useful. I would venture to say that these would be among the features which users filter there reads by most often, if not the most commonly used of all. If I run a lot of analyses, next to inputs, I think that the workflow "name" is really what I would look for first to narrow things down. And with that, I think we could indeed add an optional `workflow_name` to `RunRequest` (I think TES has something similar) - and then encourage implementers to provide a reasonable default (e.g., repo name, TRS resource name, workflow URL suffix) if a workflow name is not provided by the client. For this, `tags` feels a little too optional for me personally.

However, wouldn't strongly object to having both in `tags`.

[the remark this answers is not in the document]
wleepang: If `name` is a property of `RunSummary` I'd prefer the spect to be clear about what item is being named. If part of `RunSummary`, my expectation is that it is the name of the run and not the name of the workflow the run invoked.

With just workflow name, and assuming `tags` doesn't provide additional human readable information to help disambiguate, a list response would look like:
```json
[
    {
        "id": "e81da9ff-f5a8-4f20-9d5f-772efc4a93b4",
        "start_time": "YYYY-MM-DD HH:MM:SS Z",
        "end_time": "YYYY-MM-DD HH:MM:SS Z",
        "name": "nf-core-rnaseq",
        "tags": { ... }
    },
    {
        "id": "1ae61546-7cfa-4959-bf28-247ce1390508",
        "start_time": "YYYY-MM-DD HH:MM:SS Z",
        "end_time": "YYYY-MM-DD HH:MM:SS Z",
        "name": "nf-core-rnaseq",
        "tags": { ... }
    },
    {
        "id": "c9be874b-5b9c-48b3-879e-40986fe9dbf0",
        "start_time": "YYYY-MM-DD HH:MM:SS Z",
        "end_time": "YYYY-MM-DD HH:MM:SS Z",
        "name": "nf-core-rnaseq",
        "tags": { ... }
    },
    {
        "id": "86aaf850-b2c4-4c9d-b898-16ed3ce1daba",
        "start_time": "YYYY-MM-DD HH:MM:SS Z",
        "end_time": "YYYY-MM-DD HH:MM:SS Z",
        "name": "nf-core-rnaseq",
        "tags": { ... }
    }
]
```

This would be challenging for an end-user to find the specific run that say processed sample "LIMSJVLO6YWS". Omitting `name` from the top level of `RunSummary` and having tags explicitly for `workflow_name` and `run_name` looks like:

```json
[
    {
        "id": "e81da9ff-f5a8-4f20-9d5f-772efc4a93b4",
        "start_time": "YYYY-MM-DD HH:MM:SS Z",
        "end_time": "YYYY-MM-DD HH:MM:SS Z",
        "tags": {
            "workflow_name": "nf-core-rnaseq",
            "run_name": "LIMS2ZR5ZS29"
        }
    },
    {
        "id": "1ae61546-7cfa-4959-bf28-247ce1390508",
        "start_time": "YYYY-MM-DD HH:MM:SS Z",
        "end_time": "YYYY-MM-DD HH:MM:SS Z",
        "tags": {
            "workflow_name": "nf-core-rnaseq",
            "run_name": "LIMSXXOFXDMW"
        }
    },
    {
        "id": "c9be874b-5b9c-48b3-879e-40986fe9dbf0",
        "start_time": "YYYY-MM-DD HH:MM:SS Z",
        "end_time": "YYYY-MM-DD HH:MM:SS Z",
        "tags": {
            "workflow_name": "nf-core-rnaseq",
            "run_name": "LIMSJVLO6YWS"
        }
    },
    {
        "id": "86aaf850-b2c4-4c9d-b898-16ed3ce1daba",
        "start_time": "YYYY-MM-DD HH:MM:SS Z",
        "end_time": "YYYY-MM-DD HH:MM:SS Z",
        "tags": {
            "workflow_name": "nf-core-rnaseq",
            "run_name": "LIMS42IXEIFO"
        }
    }
]
```

IMO this provides a more useful response and maintains flexibility for clients to display or filter information how they need.

A compromise would be to maintain semantic consistency where the `name` property in RunSummary is the `run_name` and `workflow_name` is additional metadata provided in `tags`:

```json
[
    {
        "id": "e81da9ff-f5a8-4f20-9d5f-772efc4a93b4",
        "start_time": "YYYY-MM-DD HH:MM:SS Z",
        "end_time": "YYYY-MM-DD HH:MM:SS Z",
        "name": "LIMS2ZR5ZS29",
        "tags": {
            "workflow_name": "nf-core-rnaseq"
        }
    },
    {
        "id": "1ae61546-7cfa-4959-bf28-247ce1390508",
        "start_time": "YYYY-MM-DD HH:MM:SS Z",
        "end_time": "YYYY-MM-DD HH:MM:SS Z",
        "name": "LIMSXXOFXDMW",
        "tags": {
            "workflow_name": "nf-core-rnaseq"
        }
    },
    {
        "id": "c9be874b-5b9c-48b3-879e-40986fe9dbf0",
        "start_time": "YYYY-MM-DD HH:MM:SS Z",
        "end_time": "YYYY-MM-DD HH:MM:SS Z",
        "name": "LIMSJVLO6YWS",
        "tags": {
            "workflow_name": "nf-core-rnaseq"
        }
    },
    {
        "id": "86aaf850-b2c4-4c9d-b898-16ed3ce1daba",
        "start_time": "YYYY-MM-DD HH:MM:SS Z",
        "end_time": "YYYY-MM-DD HH:MM:SS Z",
        "name": "LIMS42IXEIFO",
        "tags": {
            "workflow_name": "nf-core-rnaseq"
        }
    }
]

```

uniqueg:
> I think `tags` is more important than `name` for this response. (see my [comment](https://github.com/ga4gh/workflow-execution-service-schemas/pull/207#discussion_r1165734575) above).
> 
> From a client perspective, `runid`, `(start|end)_time`, and `status` are certainly must haves. I'd say `tags` is required and it is **strongly** recommended that "run_name" and "workflow_name" tags are included.

I'm with you. It seems to me like we all agree and it's mostly a semantic argument. The main point is that we want to encourage implementers as much as possible to make this info available, without making their life unduly hard if, for whatever reason, they can't. So let it be in `tags` :)

patmagee:
@uniqueg once you update this PR to remove name and make fields like start_time and end_time optional, I will go ahead and merge it!

uniqueg:
> @uniqueg once you update this PR to remove name and make fields like start_time and end_time optional, I will go ahead and merge it!

Done

PR #213, opened 2023-09-27
Add simple run filtering on tag values

patmagee:
Currently there is no mechanism to filter runs in WES. This has been an intentional choice to ensure we are not providing an overly burdensome API to implementors. At this point there have been several common use cases which have emerged and already implemented in the community (via TES).

This PR Implements the same filtering strategy implemented by TES as part of [this](https://github.com/ga4gh/task-execution-schemas/pull/170) PR. It does not intend to be exhaustive but give a starting point to filtering workflow runs

### DIFF
--- a/openapi/workflow_execution_service.openapi.yaml
+++ b/openapi/workflow_execution_service.openapi.yaml
@@ -183,6 +183,57 @@ paths:
       description: This list should be provided in a stable ordering. (The actual ordering is implementation dependent.) When paging through the list, the client should not make assumptions about live updates, but should assume the contents of the list reflect the workflow list at the moment that the first page is requested.  To monitor a specific workflow run, use GetRunStatus or GetRunLog.
       operationId: ListRuns
       parameters:
+        - name: state
+          description: |-
+            OPTIONAL. Filter rubs by state. If unspecified,
+            no run state filtering is done.
+          in: query
+          required: false
+          schema:
+            $ref: '#/components/schemas/RunStatus'
+        - name: tag_key
+          description: |-
+            OPTIONAL. Provide key tag to filter. The field tag_key is an array
+            of key values, and will be zipped with an optional tag_value array.
+            So the query:
+            ```
+              ?tag_key=foo1&tag_value=bar1&tag_key=foo2&tag_value=bar2
+            ```
+            Should be constructed into the structure { "foo1" : "bar1", "foo2" : "bar2"}
+            
+            ```
+              ?tag_key=foo1
+            ```
+            Should be constructed into the structure {"foo1" : ""}
+            
+            If the tag_value is empty, it will be treated as matching any possible value.
+            If a tag value is provided, both the tag's key and value must be exact
+            matches for a task to be returned.
+            Filter                            Tags                          Match?
+            ----------------------------------------------------------------------
+            {"foo": "bar"}                    {"foo": "bar"}                Yes
+            {"foo": "bar"}                    {"foo": "bat"}                No
+            {"foo": ""}                       {"foo": ""}                   Yes
+            {"foo": "bar", "baz": "bat"}      {"foo": "bar", "baz": "bat"}  Yes
+            {"foo": "bar"}                    {"foo": "bar", "baz": "bat"}  Yes
+            {"foo": "bar", "baz": "bat"}      {"foo": "bar"}                No
+            {"foo": ""}                       {"foo": "bar"}                Yes
+            {"foo": ""}                       {}                            No
+          in: query
+          required: false
+          schema:
+            type: array
+            items:
+              type: string
+        - name: tag_value
+          description: |-
+            OPTIONAL. The companion value field for tag_key
+          in: query
+          required: false
+          schema:
+            type: array
+            items:
+              type: string
         - name: page_size
           in: query
           description: OPTIONAL The preferred number of workflow runs to return in a page. If not provided, the implementation should use a default page size. The implementation must not return more items than `page_size`, but it may return fewer.  Clients should not assume that if fewer than `page_size` items are returned that all items have been returned.  The availability of additional pages is indicated by the value of `next_page_token` in the response.
--- a/openapi/workflow_execution_service.openapi.yaml
+++ b/openapi/workflow_execution_service.openapi.yaml
@@ -185,7 +185,7 @@ paths:
       parameters:
         - name: state
           description: |-
-            OPTIONAL. Filter rubs by state. If unspecified,
+            OPTIONAL. Filter runs by state. If unspecified,
             no run state filtering is done.
           in: query
           required: false


uniqueg (approved):
LGTM. My only concern is that we may not have extensive experience with the tag filtering yet. @kellrott @vsmalladi: Were there any issues with the tags?

patmagee:
@uniqueg FWIW we have implemented tag filtering at DNAstack on our Wes api and have taken a slightly different approach which I think is more natural. I just suggested the current approach since it seemed that was the way tes went, but if there is no consensus on whether it is working or not we can try something else.

Basically you can specify the `tag` query param as many times as you want, representing a logical and between all of the tags.

The filter has the form `tag=(?<key>[\:\&]+)(:(?<value>.+))?`

Which then can be used like the following

```
/runs?tag=foo:bar&tag=bizz&tag=sample:3320
```

Overall we find this pretty flexible. The use case of filtering by only tag value is really not there imo, and in practice we rarely ever see someone using just a key for filtering (although not does happen).

uniqueg:
Thanks @patmagee. I like the approach you describe, but with all other things being mostly equal/equivalent, I like consistency even more :)

Less mental strain for users, devs and our future selves including similar provisions in other specs.

It's just that I never tried the TES way of implementing tag filtering myself. And so I think Kyle's and Venkat's feedback may be more valuable here.

But good to have that second option, let's see what others think.

patmagee:
@uniqueg the main problem I can see with the current approach is that there is an implicit relationship between the order of `tag_key` and `tag_value` query params.

By splitting them up you force users to think about the order that filters are supplied in, since it now matters especially when the number of tag values or keys do not match.

patmagee:
@kellrott @vsmalladi pinging you here again to see if you have any experience with how the TES filtering implementation has been received.

openapi/workflow_execution_service.openapi.yaml
```diff
@@ -183,6 +183,57 @@ paths:
       description: This list should be provided in a stable ordering. (The actual ordering is implementation dependent.) When paging through the list, the client should not make assumptions about live updates, but should assume the contents of the list reflect the workflow list at the moment that the first page is requested.  To monitor a specific workflow run, use GetRunStatus or GetRunLog.
       operationId: ListRuns
       parameters:
+        - name: state
+          description: |-
+            OPTIONAL. Filter rubs by state. If unspecified,
```
vschnei, line 188: "runs" not "rubs"

vschnei: Looks fine to me. 
@kellrott @vsmalladi  @vinjana : Any further comments  on the RP?

vinjana: I think the proposal is fine. I would, like @uniqueg, vote for consistency with associated standards (TES). 

Maybe the current way can also be extended with backwards-compatibility? E.g., `{...}` could be implicit AND, `[...]` could be implicit OR. The only problem would be NOT, but if that is solved the syntax representing a boolean algebra is complete.

For the fun of it (ignore this, if you like) I wrote some EBNF for a JSON syntax that expresses boolean expressions in a way compatible with the current proposal (= parts of `simple-and-expr` below):

```ebnf
expr ::= and-expr | or-expr | not-expr. 

// AND
and-expr ::= simple-and-expr | full-and-expr.

simple-and-expr ::= '{' ( tag-name : tag-value | expr ), ... '}'.  // i.e. non-empty JSON dict

full-and-expr ::= '{' AND ':' expr '}'.

// OR
or-expr ::= simple-or-expr | full-or-expr.

simple-or-expr ::= '[' expr, ... ']'.    // i.e. non-empty JSON list

full-or-expr ::= '{' OR ':' simple-or-expr '}'

// NOT
not-expr ::= '{' NOT ':' expr '}'.


// Operation terminals
AND ::= "and". 

OR ::= "or".

NOT ::= "not".


// Tag terminals
tag-name ::= '"' <string> '"'.

tag-value ::= '"' <string> '"'.
```

Note that AND, OR, and NOT cannot be confused with tag-name "and", "or", and "not", because in the places these operational terminals occur there is always an expression following, and all expressions start with either `{` or `[`, i.e., a JSON dict or list, which are not allowed allowed for tag-names (these have to be quoted strings). So, AFAICS this is unambiguously parseable. 

Encoding logical expressions in JSON has the advantage that no parser has to be written, but a normal JSON parser or validator can be used. Evaluating such a tree is also simple if recursion is used, I would then, by the standard, allow that the implementer implements a depth-limit for the expressions.

vschnei: Did not see EBNF  for a while now.  I guess you intended to add the EBNF to be a description for
```
 /runs:
    get:
      tags:
        - Workflow Runs
      summary: ListRuns
      description: | <HERE>
```

patmagee: @vschnei @vinjana have either of you tried the TES tag filters in the wild? The approach the @vinjana suggests points to a more complicated need that splitting the key/values into separate query params does not necessarily accomplish.

I am still skeptical about the the fact that the tag value pairs are implied by the ordinal value in the query string. It also seems prone to errors, forgetting to include one key, or one value, or adding them out of order.

If this has been well tested and it works in TES I am still favouring alignment over introducing something new. Especially if my concerns do not practically play out in the wild

vinjana: I don't have any experience with the filtering in TES. We don't have a TES as backend in the moment, but (still) include the cluster interaction directly as internal API into WESkit. 

I do agree with your reservations about using the parameter order as criterion, and I personally would also be fine with deviating from the standard in TES (somewhat contradicting my previous statement). For instance, people may use an unsorted dictionary to model those parameters, and thus will certainly run into problems if order matters. This is completely avoided by the approach you suggested: `tag={urlencode(key)}:{urlencode(value)}`.

But eventually this really is a matter of higher-level coordination and harmonization efforts in GA4GH. Not sure, whether here is the place to decide whether to prioritize "better" over "standard in other API". Maybe it also works the other way around: We deviate and find a better solution, and our solution is adopted by the others?

patmagee: @vsmalladi i would be interested in knowing your opinion here

vsmalladi: We could bring this up as discussion in the next cloud discussion for wider discussion about the solution to adopt.

kellrott:
Regarding more complex operations, one thing I would note is that a lot of the decisions in the TES spec are around simplicity of implementation on the server side. A complex set of logical operations on tag searching makes it harder for people to implement the spec on their systems. And we want as many TES servers as possible. 
If logic can be done on the client side it becomes easier it's one less thing for servers to ensure conformance to, and it's easier for code to move around. For  example, there is no good way to implement a 'AND' statement for filtering client side, because you would have to theoretically steam an unknown number of unfiltered records and to the 'AND' logic afterwards. In a large system this would be potentially 10s of thousands of more records to be retrieved, transferred and discarded. So it's something that we would want the server to do. Conversely, for an 'OR' operations, the client could send 2 requests. one for each tag, and get back a minimal set of results (a max of x2 the size of the positive set assuming every record is returned for both of the queries). So implementing the OR statement is easier to do for the client and not bother implementing on the server. 
I think the complexity of the compound logically query proposal should be evaluated using this kind of criteria.

vinjana:
@kellrott Your reasoning about the way AND and OR statements would have to be handled client side (and therefore are or are not better made server-side) makes completely sense. Thanks for sharing that, because I did not think about this before.

The argument that we need as many TES implementations as possible does not make any sense to me. The interaction with any backend cluster (SLURM, LSF, Kubernetes, ...) is orders of magnitude more complex than a logic evaluation, which anyway will usually be done by a backend database. I think the main argument really is that no more complex logics are needed by any clients.

vschnei:
Thank you all for reviewing this. I’m in favor of merging this PR.

Both @uniqueg and @vschnei have approved it, and @vinjana also expressed support while suggesting a possible future extension. Given this, I recommend that we merge the PR. We can monitor adoption, gather feedback from implementations, and iterate on a more expressive filter syntax in a future update.

@uniqueg — Could you share any experience with tag filtering in existing TES implementations?
@patmagee — Would you please fix the small typo (“rubs”)?

openapi/workflow_execution_service.openapi.yaml
patmagee:
```suggestion
            OPTIONAL. Filter runs by state. If unspecified,
```

PR #219, opened 2024-06-28
Start a list of known implementations

mr-c:


vschnei:
@mr-c Not sure if I get the purpoise of `==` instead of `--`

=== under text → creates a level 1 heading 
--- under text → creates a level 2 heading

Those are less important headings. I would stay with "--" for "known implementations" heading. 
Otherwise fine.

vschnei (approved):

Issue #176, opened 2022-02-24
Input and Output Format Specification


patmagee:
As the WES specification gains traction and there are more adopters, it is becoming apparent that interoperability between WES implementations is a real challenge. Initiatives like FASP have highlighted the numerous challenges around interacting with a given WES API. 

One of the key challenges, is the format of workflow inputs and outputs. Presently, there is no common definition for these fields in the WES specification, meaning it is up to each engine / language to dictate the format that inputs should be represented. Effectively, this requires any tooling built to interact with WES to either 1) make assumptions about the representation and only work with a specific language / engine 2) create custom logic for each implementation. This approach does not scale and does not build a strong community centred around an interoperable standard

it is clear that we need to either reduce the divergence between the different engines, or provide a solution that mitigates it.

### Inputs
workflow inputs are submitted as part of the `workflow_params` form parameter, and there is no restriction on what type of values they can represent. It is up to the user to know how to define these values for the given language they are using.

The situation is made a little bit more complex however by the fact that the `RunRequest` object implicitly expects inputs to be in a `JSON` coercible format.

### Outputs
Workflow outputs are returned by the engine as arbitrary `JSON` in the `RunLog`, therefore it is currently required that outputs must be represented as `JSON`. Since there is no imposed structure other then `JSON` each engine is free to implement the outputs section as they see fit, leading to divergence between engines.

## Proposals

I have outlined a number of proposals that immediately come to mind, however I would encourage anyone to come up with their own proposal, or ammend mine as well.

### Proposal 1: Define Abstract input and output format
The WES API should prescribe the format of all inputs and outputs in a language agnostic way. This would require all implementations to support the exact same input/output format and internally translate to whatever format is necessary for their supported languages. 

Defining a singular format is not without it's challenges. Different languages have different needs, and their input formats largely reflect that. Capturing all of the nuances into a single meta abstraction would be a challenge to do without loosing information in the process. 

Given that most (but not all) workflow languages support some sort of rudimentary type system, one approach may be to define a super set of types common across all languages. This likely is not possible, and means that the WES api would not necessarily be able to represent all types, or new types. It also implies that there are specific languages that WES supports, and any other language is out of scope.

Another possibility is to not use types at all but simply require a specific format (ie `JSON` for the inputs). This is a step better then where we were at from before, however without the ability to define HOW those `JSON` inputs are structured

#### Pros
- Allows for greater interoperability
- Provides the most please user experience when interacting with the API
[...]

Issue #173, opened 2021-10-06
Support and standard keyword for WES


jb-adams:
Currently, WES supports standard keywords for `CWL` and `WDL` workflow languages. Can we add a new `workflow_type` keyword supporting Nextflow-based workflows, perhaps `NEXTFLOW`?

uniqueg:
Just for reference:

PRs to include Nextflow and Snakemake in the corresponding TRS enum:
- Nextflow (listed as NFL): ga4gh/tool-registry-service-schemas/pull/77
- Snakemake (listed as SMK): ga4gh/tool-registry-service-schemas/pull/163 & ga4gh/tool-registry-service-schemas/pull/200

patmagee:
@caylward-ilmn this may be a case of imprecise language in the documentation. Higher up in the reference docs it also says:
>  (currently CWL or WDL formatted workflows, other types may be supported in the future)
 
Which seems to imply that CWL or WDL are the only supported languages currently

pditommaso:
Worth mentioning nextflow is already implemented by Tower API, and it's expecting `NFL` as language type for consistency with [TRS](https://github.com/pditommaso/tool-registry-service-schemas/blob/develop/src/main/resources/swagger/openapi.yaml#L741-L744)

https://tower.nf/openapi/index.html#post-/ga4gh/wes/v1/runs

jb-adams:
I agree, wasn't aware of `NFL` in TRS till now

uniqueg:
If the docs/specs are to be changed to mention supported workflow languages explicitly, then I think it would be nice if it were the same set of languages supported in TRS. From `v2.0.1` onward this will be:

- CWL
- WDL
- NFL
- Galaxy
- SMK

See [here](https://github.com/ga4gh/tool-registry-service-schemas/blob/91a57cd93caf380019d4952c0c74bb7e343e647b/openapi/openapi.yaml#L711-L716).

I am not sure whether the WES specs should include such an enumeration at all, though.

patmagee:
@uniqueg i think it definitely makes sense to keep language consistent between apis. `NFL` makes a lot of sense for this.

I am not sure what the process of getting additional languages is supported is. My guess is eventually WES will support snake make, and galaxy as well and the two lists would likely be similar.

uniqueg:
[...]

PR #226, opened 2025-07-30
Structured parameters and outputs

briandoconnor:
# Overview 

This pull request updates the Workflow Execution Service (WES) OpenAPI specification to enhance functionality in several key ways without creating breaking changes. Key changes include cleanup of our documentation, making workflow types and engines consistent in our requests/responses, creating a universal workflow parameterization passing structure (vs. relying on workflow engine specific formats), and creating a universal output format to collect outputs in a structured way regardless of workflow engine.  Claude Code was used to generate some of these changes.

**eLwazi-hosted GA4GH Hackathon**

The eLwazi hosted GA4GH hackathon 7/28-8/1 is working on this issue given the need by various groups attending the session.  For more info, see the [agenda](https://docs.google.com/document/d/1z-2Uj2R9-yLt1fTfohdUBg-kPFfxXpkWSCZz6NrACwY/edit?tab=t.0#heading=h.g7tx7vgfgncd). 

**Built Documentation**

The human-readable documentation: https://ga4gh.github.io/workflow-execution-service-schemas/preview/feature/issue-176-wes-params/docs/index.html

# More detailed description

- The major goal was to introduce a possible implementation of structured inputs and outputs for WES (issue #176)
- We wanted to do this in a non-breaking change… so this was added rather than replacing the current mechanisms (e.g. you can still use workflow_params to send a workflow-specific parameters file, you can still use multipart/form-data to upload workflow file(s) and send params).
- I added additional_workflow_urls which allows you to send multiple secondary workflow URLs beyond the primary workflow supplied by workflow_url
- In addition to multipart/form-data added the ability to submit a workflow (as workflow URLs) using application/json and a well-structured request defined in OpenAPI (rather than just strings as used in the multipart/form-data request)
- Cleaned up service-info, it was missing workflow_type and workflow_engine but these are required values when submitting a new workflow run
- Cleaned up various documentation
- Incremented version to 1.2.0
- Added 'structured_outputs' to GET /runs/{run_id} as a way of structuring outputs from a workflow in a universal way (vs. leaving it undefined as an 'outputs' object)

# Issues/questions for discussion

- For workflow_params, should this be just a string for the multipart?  Does it really need to be JSON-encoded?  That seems limiting if you're trying to pass YAML to a workflow engine…  
- Do we need "application/json" and "multipart/form-data"?  Could we get away with just "multipart/form-data"?  The problem with this is the OpenAPI is not well defined since everything is just "string".  And it hurts client generation and makes it more error prone to code against this and correctly encode request.  So there's value in having a application/json request to cleanly code against the OpenAPI for that.  But the cost of supporting two workflow run request types may be too much for implementers. 
- I noticed workflow_engine was missing from the service-info response and the workflow_engines_version array is ambiguous if this WES server supports multiple workflow engines.  So I added the former and also updated the docs to say <workflow_engine>_<workflow_engine_version> should be used for the workflow_engine_version array to make it clear when a WES server supports a variety of engines.
- Same issue for workflow_type and workflow_type_version
- I feel like both workflow_engine and workflow_type should actually be an object where we can have versions listed under each instead of two separate arrays.  But I didn't want to change this since it would be a breaking change.
- I'm not sure what "system_state_counts" in service-info is for but it's required… 
- Our list of workflow types is "CWL", "WDL", "Nextflow", or "Snakemake" (or another alternative supported by this WES instance, see service-info).  Do we want to use lowercase here (which would match supported_filesystem_protocols)?

#### Version and Logo Updates:
* Updated API version from `1.1.0` to `1.2.0`.
* Changed logo URL to reflect the latest GA4GH branding.

#### Workflow Types and Engines:
* Added support for new workflow types (`Nextflow`, `Snakemake`) in addition to `CWL` and `WDL`.
* Introduced `workflow_engine` and `workflow_engine_versions` properties to specify supported workflow engines and their versions. [[1]](diffhunk://#diff-1f171071387eb98ede480c7beb336dec6aa0a03e48a4c7a726fc19830121d804R670-R673) [[2]](diffhunk://#diff-1f171071387eb98ede480c7beb336dec6aa0a03e48a4c7a726fc19830121d804R652-R655)

#### Parameterization Enhancements:
* Added `workflow_unified_params` for universal parameter format, enabling generic parameterization across workflow types.
* Expanded descriptions for `workflow_params` and `workflow_unified_params` fields, clarifying their use cases.

#### Endpoint Descriptions:
* Enhanced descriptions for `RunWorkflow` endpoint, detailing supported content types (`application/json`, `multipart/form-data`) and file handling mechanisms.
* Improved `GetServiceInfo` endpoint description to include workflow engines and additional service information.

#### Tags and Groups:
* Added new tags (`workflowoutputs_model`, `outputobject_model`) and grouped them under `x-tagGroups` for better organization. [[1]](diffhunk://#diff-1f171071387eb98ede480c7beb336dec6aa0a03e48a4c7a726fc19830121d804R111-R118) [[2]](diffhunk://#diff-1f171071387eb98ede480c7beb336dec6aa0a03e48a4c7a726fc19830121d804R145-R153)

#### New Properties:
* Added `workflow_type` and `workflow_engine` objects to the schema for defining supported types and engines. [[1]](diffhunk://#diff-1f171071387eb98ede480c7beb336dec6aa0a03e48a4c7a726fc19830121d804R652-R655) [[2]](diffhunk://#diff-1f171071387eb98ede480c7beb336dec6aa0a03e48a4c7a726fc19830121d804R670-R673)

#### Clarifications:
* Updated descriptions for existing schema fields, ensuring consistency and clarity. [[1]](diffhunk://#diff-1f171071387eb98ede480c7beb336dec6aa0a03e48a4c7a726fc19830121d804L642-R695) [[2]](diffhunk://#diff-1f171071387eb98ede480c7beb336dec6aa0a03e48a4c7a726fc19830121d804L674-R727)

These updates make the WES API more versatile, user-friendly, and aligned with emerging standards in workflow execution.

### DIFF
--- a/openapi/workflow_execution_service.openapi.yaml
+++ b/openapi/workflow_execution_service.openapi.yaml
@@ -10,7 +10,7 @@ info:
 
     ## Executive Summary
 
-    The Workflow Execution Service (WES) API provides a standard way for users to submit workflow requests to workflow execution systems, and to monitor their execution. This API lets users run a single workflow (commonly [CWL](https://www.commonwl.org/), [WDL](http://www.openwdl.org/), [Nextflow](https://www.nextflow.io/), and [SnakeMake](https://snakemake.readthedocs.io/en/stable/) formatted workflows, WES servers describe what they support via service-info) on multiple different platforms, clouds, and environments.
+    The Workflow Execution Service (WES) API provides a standard way for users to submit workflow requests to workflow execution systems, and to monitor their execution. This API lets users run a single workflow (commonly [CWL](https://www.commonwl.org/), [WDL](http://www.openwdl.org/), [Nextflow](https://www.nextflow.io/), and [Snakemake](https://Snakemake.readthedocs.io/en/stable/) formatted workflows, WES servers describe what they support via service-info) on multiple different platforms, clouds, and environments.
 
     Key features of the API:
 
@@ -35,23 +35,23 @@ tags:
     description: Submit and monitor workflows in a WES environment
   - name: Introduction
     description: |
-      This document describes the WES API and provides details on the specific endpoints, request formats, and response.  It is intended to provide key information for developers of WES-compatible services as well as clients that will call these WES services.
+      This document describes the WES API and provides details on the specific endpoints, request formats, and responses.  It is intended to provide key information for developers of WES-compatible services as well as clients that will call these WES services.
 
       Use cases include:
       
-      - "Bring your code to the data": a researcher who has built their own custom analysis can submit it to run on a dataset owned by an external organization, instead of having to make a copy of the data
-      - Best-practices pipelines: a researcher who maintains their own controlled data environment can find useful workflows in a shared directory (e.g. Dockstore.org), and run them over their data
-      
+      - "Bring your code to the data": a researcher who has built their own custom analysis workflow can submit it to run on a dataset owned by an external organization, instead of having to download a copy of the data for local analysis.
+      - Best-practices pipelines: a researcher who maintains their own controlled data environment can find useful workflows in a shared repository (e.g. [Dockstore.org](https://dockstore.org)), and run them on their data
+
       ## Standards
         
-      The WES API specification is written in OpenAPI and embodies a RESTful service philosophy. It uses JSON in requests and responses and standard HTTP/HTTPS for information transport.
+      The WES API specification is written in OpenAPI and embodies a RESTful service philosophy. It uses JSON in requests and responses (where applicable) and standard HTTP/HTTPS for information transport.
   - name: Authorization & Authentication
     description: |
       Users must supply credentials that establish their identity and authorization in order to use a WES endpoint. We recommend that WES implementations use an OAuth2 [**bearer token**](https://oauth.net/2/bearer-tokens/), although they can choose other mechanisms if appropriate. WES callers can use the `auth_instructions_url` from the [**`service-info` endpoint**](https://ga4gh.github.io/workflow-execution-service-schemas/#/WorkflowExecutionService/GetServiceInfo) to learn how to obtain and use a bearer token for a particular implementation.
 
       The WES implementation is responsible for checking that a user is authorized to submit workflow run requests. The particular authorization policy is up to the WES implementer.
 
-      Systems like WES need to also address the ability to pass credentials with jobs for input and output access.  In the current version of WES, the passing of credentials to authenticate and authorize access to inputs and outputs, as well as mandates about necessary file transfer protocols to support, are out of scope.  However, parallel work on the GA4GH Data Repository Service is addressing ways to pass around access credentials with data object references, opening up the possibility that a future version of WES will provide concrete mechanisms for workflow runs to access data using credentials different than those used for WES.  This is a work in progress and support of DRS in WES will be added in a future release of WES.
+      Systems like WES need to also address the ability to pass credentials with jobs for input and output access.  In the current version of WES, the passing of credentials to authenticate and authorize access to inputs and outputs, as well as mandates about necessary file transfer protocols to support, are out of scope.  However, parallel work on the [GA4GH Data Repository Service](https://ga4gh.github.io/data-repository-service-schemas/) is addressing ways to pass around access credentials with data object references, opening up the possibility that a future version of WES will provide concrete mechanisms for workflow runs to access data using credentials different than those used for WES calls.  This is a work in progress and support of DRS in WES will be added in a future release of WES.
   - name: serviceinfo_model
     x-displayName: ServiceInfo
     description: |
@@ -150,7 +150,7 @@ paths:
       tags:
         - Service Info
       summary: GetServiceInfo
-      description: May include information related (but not limited to) the workflow descriptor formats, versions supported, the WES API versions supported, and information about general service availability.
+      description: Includes information related, but not limited to, the workflow descriptor formats, workflow engines, versions supported, the WES API versions supported, and information about general service availability and access.
       operationId: GetServiceInfo
       parameters: [ ]
       responses:
@@ -251,25 +251,25 @@ paths:
 
         **Content Types Supported:**
 
-        1. **`application/json`** (Recommended for single workflows submitted by URL): Submit a complete `RunRequest` object with structured data and type validation.  The advantage of this approach is the schema is explicitly defined by OpenAPI syntax (vs. just strings in multipart-form) along with workflow_unified_params object support (see below for more info).  Disadvantage is that it does not support workflow file uploads directly, which may be required for multi-file workflows.
+        1. **`application/json`** (Recommended for single workflows submitted by URL): Submit a complete `RunRequest` object with structured data and type validation.  The advantage of this approach is the schema is explicitly defined by OpenAPI syntax (vs. just strings in multipart-form) along with workflow_unified_params object support (see below for more info).  Disadvantage is that it does not support workflow file uploads directly, which may be required for multi-file workflows not available as URLs.
 
         2. **`multipart/form-data`**: Submit for cases requiring workflow file uploads (rather than just a URL). Fields are JSON-encoded strings that mirror the `RunRequest` structure.  Advantage is you can upload one or more workflow files directly while the disadvantage is you need to be careful in formating the JSON-ended string arguments correctly (multipart form submissions only support strings hence the limitation). 
 
         **File Handling:**
 
-        - **JSON requests**: Workflow file is referenced by URL or path. Use `workflow_unified_params` with File objects for rich metadata and generic parameterization.
+        - **`application/json` requests**: Workflow file is referenced by URL or path. Use `workflow_unified_params` with File objects for rich metadata and generic parameterization.
 
-        - **Multipart requests**: The `workflow_attachment` array uploads files required for workflow execution. Files are staged to a temporary directory with Content-Disposition headers containing filenames. Subdirectories are allowed but parent directory references ('..') are prohibited for security purposes.
+        - **`multipart/form-data` requests**: The `workflow_attachment` array uploads files required for workflow execution. Files are staged to a temporary directory with Content-Disposition headers containing filenames. Subdirectories are allowed but parent directory references ('..') are prohibited for security purposes.
 
         **Parameter Formats:**
 
-        - **`workflow_params`**: Language-specific parameters (CWL, WDL, Nextflow, SnakeMake format).  This field is used to send the native parameterization file (encoded as a string) used by the workflow engine.
+        - **`workflow_params`**: Language-specific parameters (CWL, WDL, Nextflow, Snakemake format).  This field is used to send the native parameterization file (encoded as a string) used by the workflow engine.
 
         - **`workflow_unified_params`**: Universal parameter format that WES server implementation converts to the target workflow language.  This allows for a more generic way to pass parameters that can be converted to the appropriate native format for the specified workflow_type.  This is an optional field, but if provided, it takes precedence over `workflow_params`.
 
         **Required Fields:**
 
-        - `workflow_type`: Workflow language ("CWL", "WDL", "Nextflow", "SnakeMake", or another alternative supported by this WES instance, see service-info)
+        - `workflow_type`: Workflow language ("CWL", "WDL", "Nextflow", "Snakemake", or another alternative supported by this WES instance, see service-info)
 
         - `workflow_type_version`: The workflow descriptor type version, must be one supported by this WES instance
 
@@ -296,7 +296,7 @@ paths:
                   description: JSON-encoded unified workflow parameters (see RunRequest for how to encode)
                 workflow_type:
                   type: string
-                  description: Workflow descriptor type (CWL, WDL, Nextflow, SnakeMake, etc. see service-info for supported types)
+                  description: Workflow descriptor type must be "CWL", "WDL", "Nextflow", or "Snakemake" currently (or another alternative supported by this WES instance, see service-info)
                 workflow_type_version:
                   type: string
                   description: The workflow descriptor type version, must be one supported by this WES instance
@@ -662,6 +662,10 @@ components:
               items:
                 type: string
               description: The filesystem protocols supported by this service, currently these may include common protocols using the terms 'http', 'https', 'sftp', 's3', 'gs', 'file', or 'synapse', but others  are possible and the terms beyond these core protocols are currently not fixed.   This section reports those protocols (either common or not) supported by this WES service.
+            workflow_engine:
+              type: object
+              additionalProperties:
+                $ref: '#/components/schemas/WorkflowEngine'
             workflow_engine_versions:
               type: object
               additionalProperties:
@@ -683,7 +687,7 @@ components:
               type: object
               additionalProperties:
                 type: string
-                description: A message containing useful information about the running service, including supported versions and default settings.
+                description: Tags containing useful information about the WES service.
           required:
             - workflow_type_versions
             - supported_wes_versions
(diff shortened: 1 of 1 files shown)


openapi/workflow_execution_service.openapi.yaml
```diff
@@ -235,56 +246,90 @@ paths:
...
+                  description: JSON-encoded array of additional workflow URLs (see RunRequest for how to encode)
                 workflow_attachment:
                   type: array
                   items:
                     type: string
                     format: binary
-                  description: ''
-        required: false
+                  description: >-
+                    Files to be staged for workflow execution. You set the filename/path using the Content-Disposition header in the multipart form submission. For example 'Content-Disposition: form-data; name="workflow_attachment"; filename="workflows/helper.wdl"'. The files are staged to a temporary directory and can be referenced by relative path in the workflow_url.
```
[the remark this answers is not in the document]
briandoconnor: Fixed

openapi/workflow_execution_service.openapi.yaml
```diff
@@ -742,45 +795,158 @@ components:
...
+                        items:
+                          description: Allowed value (type matches parameter type)
+                        description: List of allowed values
+                      min_items:
+                        type: integer
+                        description: Minimum number of items (for Array types)
+                      max_items:
+                        type: integer
+                        description: Maximum number of items (for Array types)
+                required:
```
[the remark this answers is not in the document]
briandoconnor: I don't think this is right...

pvanheus:
Just dropping in this from @suecharo / @inutano - Sapporo extended WES in a few ways - https://petstore.swagger.io/?url=https://raw.githubusercontent.com/sapporo-wes/sapporo-service/main/sapporo-wes-spec-2.0.0.yml 

One thing that is interesting in their extensions is adding workflow_attachments_obj, which makes it, I think, somewhat easier to specify the destination filename of attachments. Perhaps going even further would be useful and specifying them are File or Directory types.

openapi/workflow_execution_service.openapi.yaml
```diff
@@ -742,45 +795,159 @@ components:
...
+                  description:
+                    type: string
+                    description: Human-readable parameter description
+                  optional:
+                    type: boolean
+                    default: false
+                    description: Whether this parameter is optional
+                  default:
+                    description: Default value if parameter not provided (type depends on 'type' field)
+                  constraints:
```
vinjana, line 860: I'm wondering about the use of uploading validation constraints when submitting a workflow run.

I totally agree that setting up validation-constraints in a WES makes sense, if the WES is installing a new workflow. However, this implementation has some disadvantages

1. **Consistency**: Defining such constraints when requesting a workflow **run** allows that the client define these constraints differently for different runs, or that different clients (users) use different constraints. 
2. **Redundancy**: The clients probably would have to provide these ~parameters~ constraints every time they submit a run, and it would always be the same ~parameters~ constraints, right?
3. **Competencies**: Finally, the user of the WES instance is not necessarily the same person (or expert level) as the administrator -- or whoever is responsible for installing workflows. The current implementation puts the load of defining these constraints on the clients, although it should better be put on some kind of an administrator. This is particularly relevant with human data and tight security constraints that may involve workflow auditing, and also is problematic if there are a **lot** of clients.

These problems may hint at that we are not modelling the domain in sufficient detail.

## Some thoughts in this direction

Logically, to me the validation constraints belong with the workflow, not with the workflow run.

Think of a use case like this:

1. Some client application first requests a new workflow. This could be a a data user but also a WES administrator or somebody involved in governance of the instance. Think of security auditing etc. The workflow installation route could have separate authentication.
4. The workflow is installed. Dependent on the workflow this may involve a lot of steps, such as downloading containers, running integration tests to verify the installation. (Of course, ideally just some containers would be pulled).
5. The clients use the workflow in the usual way.


We currently don't have a separate endpoint for requesting the installation of a new workflow, e.g., for downloading from a TRS. I'm also not saying we need it -- there is value in keeping the API simple -- however, I just wanted to make the point, that we _could_ have such an endpoint and that also that would have some valid use case. An installation of a workflow could be considered a "resource" - in the REST sense.

openapi/workflow_execution_service.openapi.yaml
```diff
@@ -742,45 +795,159 @@ components:
...
+            OPTIONAL
+
+            The workflow engine version, must be one supported by this WES instance (see service-info for supported engine versions). If workflow_engine is provided, but workflow_engine_version is not, servers can make no assumptions with regard to the engine version the WES instance uses to process the request if that WES instance supports multiple versions of the requested engine.
         workflow_url:
           type: string
           description: >-
             REQUIRED
 
-            The workflow CWL or WDL document. When `workflow_attachments` is used to attach files, the `workflow_url` may be a relative path to one of the attachments.
+            The primary workflow document. When `workflow_attachments` is used to attach files, the `workflow_url` may be a relative path to one of the attachments that is the primary workflow to be executed.
```
vinjana, line 932: The comment is not limiting the use case in the moment ("may be relative"), but it also does not reflect the whole flexibility of the approach, which _at least_ includes the following:

* files uploaded as attachment (e.g., `file:main.nf`)
* TRS URIs to files downloaded from TRS server (e.g., `trs://server:port/path/to/package`)
* files in globally shared (at least for the user group) central workflow installation (e.g., `file:/workflows/AwesomeWorkflow/main.nf`).

You might want to cover more of use cases here -- also because this YAML is the main documentation, and a more inclusive explanation may help the implementers to understand the standard better.

openapi/workflow_execution_service.openapi.yaml
```diff
@@ -942,8 +1122,18 @@ components:
...
       description: Available workflow engine versions supported by a given instance of the service.
+    WorkflowEngine:
+      title: WorkflowEngine
+      type: object
+      properties:
+        workflow_engine:
+          type: array
+          items:
+            type: string
+          description: An array of one or more acceptable workflow engines. Since a server may support multiple engines and version, the recommendation is to encode the workflow_engine_version array as `<workflow_engine>_<version>` where `workflow_engine` values match this array for clarity.
```
vinjana, line 1135:
```suggestion
          description: An array of one or more acceptable workflow engines. If the server supports multiple engine versions, encode the versions in `workflow_engine_versions`.
```

1. Better not repeat the documentation of `workflow_engine_version` here, because it distracts from the meaning of `workflow_engine` and is harder to maintain (DRY principle).
2. I still think that the reference to `workflow_engine_version` is useful.

openapi/workflow_execution_service.openapi.yaml
```diff
@@ -975,8 +1165,84 @@ components:
...
+        version:
+          type: string
+          description: Version of the structured outputs format
+          default: "1.0"
+        outputs:
+          type: object
+          description: Named workflow outputs with structured metadata
+          additionalProperties:
+            $ref: '#/components/schemas/OutputObject'
+    OutputObject:
```
vinjana, line 1182: Do I understand correctly, that this is either a single value or an array of `OutputObject`. So this would be a recursive definition, right? This looks quite powerful. Nice!

openapi/workflow_execution_service.openapi.yaml
```diff
@@ -975,8 +1165,84 @@ components:
...
+        size:
+          type: integer
+          format: int64
+          description: Size in bytes (for File/Directory class)
+        checksum:
+          type: string
+          description: File integrity hash in format 'algorithm:hash' (e.g., 'sha256:abc123...')
+        format:
+          type: string
+          description: MIME type or format identifier (e.g., EDAM ontology reference)
```
vinjana, line 1208: Maybe it would be useful to directly fix or at least suggest certain aspects of the `format`. Maybe the best solution is to at least suggest something like URIs for terms, e.g. MIME types (https://www.iana.org/assignments/media-types/media-types.xhtml), and that multiple terms can be used, separated by commata -- or directly make this field a `List[String]` field.

In the extreme you could define this as of `type` `List[URI]`.

openapi/workflow_execution_service.openapi.yaml
```diff
@@ -975,8 +1165,84 @@ components:
...
+      properties:
+        class:
+          type: string
+          enum: ["File", "Directory", "Array", "String", "Integer", "Float", "Boolean"]
+          description: Type of the output object
+        value:
+          description: Output value (type depends on class field)
+        location:
+          type: string
+          description: Absolute path or URL to the output file/directory (for File/Directory class)
```
vinjana, line 1195: Why absolute path? Why a path at all? Should this not better be a URI? Consider a WES that makes final files available via S3 buckets.

I guess this should **NOT** include the basename? Or should it?

openapi/workflow_execution_service.openapi.yaml
```diff
@@ -975,8 +1165,84 @@ components:
...
+              description: Unique identifier of the generating task
+            command:
+              type: array
+              items:
+                type: string
+              description: Command line that generated this output
+            description:
+              type: string
+              description: Human-readable description of the output
+        secondary_files:
```
vinjana, line 1235: I think the WES implementer or administrator will often not be able to decide what is primary or secondary. There may be *many* workflows and a WES instance may even promiscuously allow to run arbitrary workflows downloaded from a TRS, right? Therefore, to be useful this field will usually rely on the output the workflow.

Therefore, I would question whether this adds much value to the API. At most it will be used to hand through information that is anyways available in some workflow output file. But then the client/data user can access the workflow result files and obtains the information from there.

BTW: There may even be cases, where the ideas of the **workflow implementer** of what constitutes a primary or secondary file might not concord with what the user thinks.

openapi/workflow_execution_service.openapi.yaml
```diff
@@ -235,56 +246,90 @@ paths:
...
               properties:
                 workflow_params:
                   type: string
+                  description: JSON-encoded native workflow parameters (see RunRequest for how to encode)
+                workflow_unified_params:
+                  type: string
+                  description: JSON-encoded universal workflow parameters (see RunRequest for how to encode)
                 workflow_type:
                   type: string
+                  description: Workflow descriptor type must be "CWL", "WDL", "Nextflow", or "Snakemake" currently (or another alternative supported by this WES instance, see service-info)
```
vschnei, line 299: TRS (from 2.0.1 on ) is using :

- CWL
- WDL
- NFL
- Galaxy
- SMK

We should stick to that as well. See discussion https://github.com/ga4gh/workflow-execution-service-schemas/issues/173

openapi/workflow_execution_service.openapi.yaml
```diff
@@ -742,45 +795,159 @@ components:
...
+            The workflow run parameterizations (JSON encoded), including input and output file locations.
+            Either `workflow_params` or `workflow_unified_params` must be provided, `workflow_unified_params` takes precedence.
+        workflow_unified_params:
+          type: object
+          description: >-
+            OPTIONAL
+
+            Unified parameter format that can be converted to workflow-language-specific format.
+            If provided, takes precedence over workflow_params. WES implementations should
+            convert these to the appropriate native format for the specified workflow_type.
```
vschnei, line 809: Please don’t get me wrong—I’m not opposed to using `workflow_unified_params`. However, its impact is limited to a relatively small subset of parameters, primarily those related to compute resources (e.g., `cores`, `memory`). Other parameters, such as `repeat`, require special handling since their behavior varies depending on the engine.

As @vinjana suggested in our discussion, we might consider introducing a parameter property like `wes_version `to ensure backward compatibility for these unified parameters.

openapi/workflow_execution_service.openapi.yaml
```diff
@@ -742,45 +795,159 @@ components:
...
           additionalProperties:
             type: string
+          description: >-
+            OPTIONAL
+
+            Arbitrary key-value tags to associate with the workflow run for organization, tracking, or metadata purposes. Examples include experiment names, user identifiers, sample IDs, or project codes.
         workflow_engine_parameters:
           type: object
           additionalProperties:
             type: string
```
vschnei, line 914: description is missing.

JSON-encoded engine-specific parameters

vschnei (changes requested):

openapi/workflow_execution_service.openapi.yaml
vinjana: Is `workflow_unified_params` not related to `workflow_params` rather than `workflow_engine_params`? At least the comments suggest that `workflow_params` and `workflow_unified_params` belong together.

But I understand the confusion, because also for workflow engine parameters one may think of standardizing them in a similar way, with ontology terms, etc.

openapi/workflow_execution_service.openapi.yaml
kraatz: I agree, that it seems redundant to pass the parameter type for every workflow run. The parameter types are not changing, so it makes sense to define them with the workflow or even with the engine?

vinjana: I guess, if you are talking about workflow parameters (e.g., bwa-mem parameters) with the workflow.

If you are talking about workflow engine parameters (e.g., number of cores of the control process or whether to rerun (and complement) a previous run), then with the engine. 

In both cases, the problem is analogous. Ideally, a similar solution would be found. I think something like EDAM can be helpful in both cases.

openapi/workflow_execution_service.openapi.yaml
patmagee: Probably should be defined as an enum lower down

openapi/workflow_execution_service.openapi.yaml
```diff
@@ -235,56 +246,90 @@ paths:
...
+                  description: The workflow engine, must be one supported by this WES instance. Required if workflow_engine_version is provided.
                 workflow_engine_version:
                   type: string
+                  description: The workflow engine version, must be one supported by this WES instance. If workflow_engine is provided, but workflow_engine_version is not, servers can make no assumptions with regard to the engine version the WES instance uses to process the request if that WES instance supports multiple versions of the requested engine.
                 workflow_engine_parameters:
                   type: string
+                  description: JSON-encoded engine-specific parameters (see RunRequest for how to encode)
                 workflow_url:
                   type: string
+                  description: The workflow document. When workflow_attachment is used to attach files, the workflow_url may be a relative path to one of the attachments.
```
patmagee, line 317:
```suggestion
                  description: The path to the workflow document. When workflow_attachment is used to attach files, the workflow_url may be a relative path to one of the attachments.
```

openapi/workflow_execution_service.openapi.yaml
```diff
@@ -235,56 +246,90 @@ paths:
...
+                  description: JSON-encoded array of additional workflow URLs (see RunRequest for how to encode)
                 workflow_attachment:
                   type: array
                   items:
                     type: string
                     format: binary
-                  description: ''
-        required: false
+                  description: >-
+                    Files to be staged for workflow execution. You set the filename/path using the Content-Disposition header in the multipart form submission. For example 'Content-Disposition: form-data; name="workflow_attachment"; filename="workflows/helper.wdl"'. The files are staged to a temporary directory and can be referenced by relative path in the workflow_url.
```
patmagee, line 327: Looking in the RFC for filenames, it does seem like it is strongly encouraged to NOT include folder structures in the `filename` param.

openapi/workflow_execution_service.openapi.yaml
patmagee: IMO constraints are typically properties of a workflow in the case of a WES. Putting their definition on the end user feels needless. 

If the user is setting these constraints themselves, why not just modify the inputs to match their constraints before submitting the workflow? Where I have heard the request for constraints come from is workflow authors who want to define things like ranges of values, or enums.

Constraints would probably be better living along side TRS or directly within the workflow itself (if the language supported it)

openapi/workflow_execution_service.openapi.yaml
```diff
@@ -742,45 +795,159 @@ components:
...
+            Either `workflow_params` or `workflow_unified_params` must be provided, `workflow_unified_params` takes precedence.
+        workflow_unified_params:
+          type: object
+          description: >-
+            OPTIONAL
+
+            Unified parameter format that can be converted to workflow-language-specific format.
+            If provided, takes precedence over workflow_params. WES implementations should
+            convert these to the appropriate native format for the specified workflow_type.
+          properties:
```
patmagee, line 810: These should probably be moved into a separate object: `WorkflowUnifiedParams`

Issue #156, opened 2020-08-17
Updates now that GA4GH AAI standard is confirmed


briandoconnor:
This is a ticket that corresponds to ticket [#294](https://github.com/ga4gh/data-repository-service-schemas/issues/294) on the DRS repo.  The goal is to make updates to both the spec and openapi schema to accommodate [GA4GH passports](https://github.com/ga4gh-duri/ga4gh-duri.github.io/blob/master/researcher_ids/ga4gh_passport_v1.md#passport).

briandoconnor:
We're looking at this as part of the eLwazi Hackathon July 2025.

I see the following related tickets:
- https://github.com/ga4gh/workflow-execution-service-schemas/issues/18
- https://github.com/ga4gh/workflow-execution-service-schemas/issues/131 
- https://github.com/ga4gh/workflow-execution-service-schemas/issues/156

Issue #131, opened 2019-05-06
Add a `workflow_params_drs_headers` field to `RunWorkflow` endpoint


jaeddy:
[Notes](https://docs.google.com/document/d/1XsYCMG2JkJen-ZjPAcy7c9RfA41wbzAmXBkB3iQm6PI/edit?usp=sharing) from GA4GH Connect Implementation meeting in Hinxton (April 2019):

> - it’s an array of {<DRS server>, <headers>} pairs
> - WES is expected, whenever fetching a DRS URI, to use the specified headers based on which server is specified in that DRS URI

I'd like to figure out a way that this field wouldn't be quite so DRS specific/dependent... but need to give it some more thought.

briandoconnor:
James asked if we can make this less DRS-specific...

briandoconnor:
We're looking at this as part of the eLwazi Hackathon July 2025.

I see the following related tickets:
- https://github.com/ga4gh/workflow-execution-service-schemas/issues/18
- https://github.com/ga4gh/workflow-execution-service-schemas/issues/131 
- https://github.com/ga4gh/workflow-execution-service-schemas/issues/156

Issue #18, opened 2018-01-18
Data access credentials


david4096:
If the execution environment needs to access data using API keys, can we offer an interoperable way to achieve this? Are these workflow parameters to the workflow?

brucehoff:
One approach is to use an OAuth approach in which the data source is the "authorization/resource server" and the execution environment is the OAuth *client*.  The workflow initiator would obtain an [authorization code](https://www.oauth.com/oauth2-servers/access-tokens/authorization-code-request/) for the client (execution environment) to access the necessary resources.  This code could be passed *in the clear* as a workflow parameter because it doesn't provide access to anything itself.  The execution environment then uses the access code along with its credentials to obtain an access token from the authorization server which lets it, say, download a file.  

I would *not* pass an access token as a workflow parameter because it's not secure.

The set up for the secure approach involves establishing the execution environment as an OAuth client to the authorization server whose resources are to be accessed.  Further, the WES client initiating the authorization code request needs to know the OAuth client ID of the execution environment.

ruchim:
This sounds like a good suggestion. There was a ticket that Brian recently opened: https://github.com/ga4gh/workflow-execution-service-schemas/issues/156

The new ticket has recommendations to align with the latest GA4GH AAI standard, which gets to the heart of managing access as well as best practices on managing credentials. I recommend closing this ticket and focusing on the other instead. @david4096 @brucehoff let me know if you have concerns with this approach.

brucehoff:
@ruchim you ask

>  I recommend closing this ticket and focusing on the other instead. @david4096 @brucehoff let me know if you have concerns with this approach.

Yes:  The [ticket you linked](#156)  just says, "The goal is to make updates to both the spec and openapi schema to accommodate GA4GH passports."  There's no assurance that solving that issue will solve this one.  In particular, a GA4GH passport (as I understand it) is a form of OAuth identity token, not an access token or authorization code. So I don't see how it would address this issue.  Can you elaborate?

ruchim:
Just sharing some context into how I'm approaching Data Access for WES.

I've heard 3 primary use cases here (a few inspired by other GA4GH standards)
1. I'm a user that has access data in an S3 bucket and would like to access the cloud native object as an input to my WES workload.
2. I'm a user that has a DRS URI and would like to access the underlying file it references as an input to my WES workload.
[...]

PR #227, opened 2025-07-31
Data Access Credentials

briandoconnor:
# Overview 

This pull request updates the Workflow Execution Service (WES) OpenAPI specification (starting with #226) to enhance functionality in several key ways without creating breaking changes (hopefully). Key changes include 1) adding POST endpoints to support authentication/authorizing with a GA4GH Passport rather than a bearer token, 2) providing credentials for a WES server to access DRS (or other) inputs, and 3) for providing back credentials for the output of the workflow run to be accessible.  Claude Code was used to generate some of these changes.

**Related issues**

- https://github.com/ga4gh/workflow-execution-service-schemas/issues/18
- https://github.com/ga4gh/workflow-execution-service-schemas/issues/131 
- https://github.com/ga4gh/workflow-execution-service-schemas/issues/156

 **Related Standards** 

This implementation aligns with:

- [GA4GH Passport Specification](https://github.com/ga4gh-duri/ga4gh-duri.github.io/blob/master/researcher_ids/ga4gh_passport_v1.md)
- [GA4GH DRS Specification](https://ga4gh.github.io/data-repository-service-schemas/) for
  data access patterns
- [WES 1.1.0 Specification](https://ga4gh.github.io/workflow-execution-service-schemas/docs/)
  for backward compatibility

**Related PR**

The feature branch behind this PR is based on `feature/issue-176-wes-params` and not `develop` since I was building off of the structured input/output.  See #226 

**eLwazi-hosted GA4GH Hackathon**

The eLwazi hosted GA4GH hackathon 7/28-8/1 is working on this issue given the need by various groups attending the session.  For more info, see the [agenda](https://docs.google.com/document/d/1z-2Uj2R9-yLt1fTfohdUBg-kPFfxXpkWSCZz6NrACwY/edit?tab=t.0#heading=h.g7tx7vgfgncd). 

**Built Documentation**

The human-readable documentation: https://ga4gh.github.io/workflow-execution-service-schemas/preview/feature/issue-18-data-access-creds/docs/index.html

## Issues/questions for discussion
- See security and implementation sections below...
- Do passport visas need to evolve to be useful for using passports as an AuthN/Z token for making WES requests?  Right now, I know of Passports being used with visas that specifically talk about data access but the changes made in this PR open up using Passports for compute.
- Is it better to have a `POST /runs/list` endpoint instead of overloading `POST /runs` based on what request body schema is used?
- In previous conversations, the idea of delegating tokens to a WES service brought up security concerns.  I think it's a valid point however we see examples with Driver projects of Passports being used in this way... retrieved by one service from the passport broken and handed to another service for data access.  So it's a concern but I think we need to weigh the positives with negatives here and realize one security profile doesn't meet every implementers needs.
- do we need to make it more explicit in the schema that supporting passports is optional?  A service may choose to not implement passports at all and they are still a valid WES.  Do we need an error code.  For example:
  - 405 Method Not Allowed - The endpoint exists, but the HTTP method (GET, POST, etc.) is
  not supported... useful for the POST endpoints on existing paths.
  - 501 Not Implemented - The endpoint is defined but not yet implemented by the server. Maybe this is the response for POST /runs when trying to list runs with a passport when that's not implemented?
- Do we need something in service-info that says this server supports POSTs w/ Passports or not?
- Do we need something in service-info to say the WES server supports delegated credentials? And more generally, the structured inputs/outputs.
- What's required?  Does every WES 1.2 server need to support DRS+HTTP URLs with auth for inputs?  For outputs the server isn't really mandated to use DRS or even HTTP outputs.
- in this PR, outputs are deposited (potentially) in a location and credentials (if applicable) are handed back to the WES caller.  However, do we want a mode additionally where the submitter tells WES where they want the outputs to be written to (along with credentials to push data there)?

## Authentication/Authorization for WES Requests 

I followed the same pattern here as used in [DRS](https://ga4gh.github.io/data-repository-service-schemas/preview/develop/docs/), making POST methods for existing endpoints that now support including a passport array in the body.  These mirror the existing GET endpoints that use a bearer token.

### Making Requests

Here's an example showing a bearer token request vs. the Passport in body request:

```
#  Bearer Token (existing):
  curl -H "Authorization: Bearer <token>" https://wes.example.com/ga4gh/wes/v1/runs

#  GA4GH Passport (new):
  curl -X POST https://wes.example.com/ga4gh/wes/v1/runs/passport \
    -H "Content-Type: application/json" \
    -d '{"passport": ["<jwt1>", "<jwt2>"]}'
```

### POST and /runs endpoint 

This approach makes the `/runs` endpoint complicated.  It feels overloaded since it's doing multiple things:

- if you do a `GET /runs` with a bearer token it's a list of runs
- if you do a `POST /runs` with a bearer token (or Passport in the body) and either `application/json` body schema of type `PassportRunRequest` or `multipart/form-data` you get a new workflow run
- if you do a `POST /runs` with a passport in the body using `application/json` body schema of type `RunListRequest` then you get a list of runs 

So it might be more clear if we have a `POST /runs/list` endpoint instead... so `/runs` is less overloaded.

## Sending Credentials

In addition to adding support for Passports in WES requests (as an alternative to bearer tokens), this PR includes syntax extensions to allow for providing credentials to access inputs and pass back credentials so callers can access outputs.

### Data Inputs

Here's how a client would structure a workflow submission:

```
  {
    "workflow_type": "WDL",
    "workflow_type_version": "1.0",
    "workflow_url": "https://example.com/workflow.wdl",

    "credentials": {
      "tcga_passport": {
        "type": "passport",
        "passport": ["eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1N..."]
      },
      "s3_token": {
        "type": "bearer_token",
        "token": "aws-token-here..."
      },
      "api_access": {
        "type": "api_key",
        "key": "abc123",
        "header_name": "X-Custom-Key"
      }
    },

    "workflow_unified_params": {
      "version": "1.0",
      "parameters": {
        "input_bam": {
          "type": "File",
          "value": "drs://drs.example.org/abc123",
          "file_metadata": {
            "credential_id": "tcga_passport",
            "size": 1000000,
            "checksum": "sha256:abc123..."
          }
        },
        "reference_genome": {
          "type": "File",
          "value": "s3://bucket/reference.fa",
          "file_metadata": {
            "credential_id": "s3_token"
          }
        },
        "public_annotation": {
          "type": "File",
          "value": "https://public.example.com/annotation.bed"
          // No credential_id needed for public files
        }
      }
    }
  }

```

### Data Outputs

Now GetRunLog responses can include credentials for accessing outputs:

```
{
    "run_id": "workflow-123",
    "state": "COMPLETE",

    "output_credentials": {
      "s3_output_access": {
        "type": "bearer_token",
        "token": "aws-s3-token-456..."
      },
      "restricted_results": {
        "type": "passport",
        "passport": ["eyJ0eXAiOiJKV1QiLCJhbGci..."]
      }
    },

    "structured_outputs": {
      "version": "1.0",
      "outputs": {
        "result_bam": {
          "class": "File",
          "location": "s3://results-bucket/output.bam",
          "credential_id": "s3_output_access",
          "size": 2000000,
          "checksum": "sha256:def456..."
        },
        "analysis_report": {
          "class": "File",
          "location": "drs://restricted.example.org/report123",
          "credential_id": "restricted_results",
          "format": "application/pdf"
        },
        "public_summary": {
          "class": "File",
          "location": "https://public.example.com/summary.txt"
          // No credential_id needed for public files
        }
      }
    }
  }

```

## Security Considerations

- Credentials are only included in responses when the requesting client is authorized to
  access the outputs.  We need to really think about how this works for Passports since typically these are issued through an OIDC flow and not just handed around.  API keys and bearer tokens are a little easier since it would be possible to issue temporary/limited scoped tokens.
- Large passport JWTs are defined once and referenced multiple times to minimize request/response size... this should help with performance issues (less a security concern)
- Credential scope should be limited to specific file resources, not broad system access... helps to address the concern about passing around powerful tokens.  If possible, Passports passed in for inputs access could be downscoped to just include visas relevant to the files in the inputs.
- It's unclear if WES servers should validate passport signatures and visa claims before using the passports for inputs access.  I suspect not but something to discuss
- This implementation assumes you can pass your passport/token/api key to another system and have that system use it on your behalf.  This would work fine in the DRS world except if there's an additional trust layer implemented (as some Driver Projects are discussing).  In this case, the WES server might not be trusted by the DRS server and, even with having a correct Passport, might be denied access to files.  This is a fluid topic, we don't have a clear trust mechanism but if trust mechanisms do get implemented (either in the specs or outside) it could affect the model presented in this PR of being able to pass around credentials.

## Implementation Notes

- **Passport Support**: WES servers MAY implement passport authentication (not required)
- **Credential Types**: Servers SHOULD support at least bearer tokens, MAY support
  passport credentials
- **Delegated Credentials**: Servers MUST support DRS and HTTP URLs for delegated credentials
- **Backward Compatibility**: All existing authentication methods MUST continue to work.  The current inputs/multipart/form-data and output approaches need to continue to work.  So a WES 1.1 client should be able to work with a WES 1.2 server without modification.
- **Error Handling**: Servers should return appropriate HTTP status codes for credential
  failures.

### DIFF
--- a/openapi/workflow_execution_service.openapi.yaml
+++ b/openapi/workflow_execution_service.openapi.yaml
@@ -10,7 +10,7 @@ info:
 
     ## Executive Summary
 
-    The Workflow Execution Service (WES) API provides a standard way for users to submit workflow requests to workflow execution systems, and to monitor their execution. This API lets users run a single workflow (commonly [CWL](https://www.commonwl.org/), [WDL](http://www.openwdl.org/), [Nextflow](https://www.nextflow.io/), and [SnakeMake](https://snakemake.readthedocs.io/en/stable/) formatted workflows, WES servers describe what they support via service-info) on multiple different platforms, clouds, and environments.
+    The Workflow Execution Service (WES) API provides a standard way for users to submit workflow requests to workflow execution systems, and to monitor their execution. This API lets users run a single workflow (commonly [CWL](https://www.commonwl.org/), [WDL](http://www.openwdl.org/), [Nextflow](https://www.nextflow.io/), and [Snakemake](https://Snakemake.readthedocs.io/en/stable/) formatted workflows, WES servers describe what they support via service-info) on multiple different platforms, clouds, and environments.
 
     Key features of the API:
 
@@ -35,23 +35,23 @@ tags:
     description: Submit and monitor workflows in a WES environment
   - name: Introduction
     description: |
-      This document describes the WES API and provides details on the specific endpoints, request formats, and response.  It is intended to provide key information for developers of WES-compatible services as well as clients that will call these WES services.
+      This document describes the WES API and provides details on the specific endpoints, request formats, and responses.  It is intended to provide key information for developers of WES-compatible services as well as clients that will call these WES services.
 
       Use cases include:
       
-      - "Bring your code to the data": a researcher who has built their own custom analysis can submit it to run on a dataset owned by an external organization, instead of having to make a copy of the data
-      - Best-practices pipelines: a researcher who maintains their own controlled data environment can find useful workflows in a shared directory (e.g. Dockstore.org), and run them over their data
-      
+      - "Bring your code to the data": a researcher who has built their own custom analysis workflow can submit it to run on a dataset owned by an external organization, instead of having to download a copy of the data for local analysis.
+      - Best-practices pipelines: a researcher who maintains their own controlled data environment can find useful workflows in a shared repository (e.g. [Dockstore.org](https://dockstore.org)), and run them on their data
+
       ## Standards
         
-      The WES API specification is written in OpenAPI and embodies a RESTful service philosophy. It uses JSON in requests and responses and standard HTTP/HTTPS for information transport.
+      The WES API specification is written in OpenAPI and embodies a RESTful service philosophy. It uses JSON in requests and responses (where applicable) and standard HTTP/HTTPS for information transport.
   - name: Authorization & Authentication
     description: |
       Users must supply credentials that establish their identity and authorization in order to use a WES endpoint. We recommend that WES implementations use an OAuth2 [**bearer token**](https://oauth.net/2/bearer-tokens/), although they can choose other mechanisms if appropriate. WES callers can use the `auth_instructions_url` from the [**`service-info` endpoint**](https://ga4gh.github.io/workflow-execution-service-schemas/#/WorkflowExecutionService/GetServiceInfo) to learn how to obtain and use a bearer token for a particular implementation.
 
       The WES implementation is responsible for checking that a user is authorized to submit workflow run requests. The particular authorization policy is up to the WES implementer.
 
-      Systems like WES need to also address the ability to pass credentials with jobs for input and output access.  In the current version of WES, the passing of credentials to authenticate and authorize access to inputs and outputs, as well as mandates about necessary file transfer protocols to support, are out of scope.  However, parallel work on the GA4GH Data Repository Service is addressing ways to pass around access credentials with data object references, opening up the possibility that a future version of WES will provide concrete mechanisms for workflow runs to access data using credentials different than those used for WES.  This is a work in progress and support of DRS in WES will be added in a future release of WES.
+      Systems like WES need to also address the ability to pass credentials with jobs for input and output access.  In the current version of WES, the passing of credentials to authenticate and authorize access to inputs and outputs, as well as mandates about necessary file transfer protocols to support, are out of scope.  However, parallel work on the [GA4GH Data Repository Service](https://ga4gh.github.io/data-repository-service-schemas/) is addressing ways to pass around access credentials with data object references, opening up the possibility that a future version of WES will provide concrete mechanisms for workflow runs to access data using credentials different than those used for WES calls.  This is a work in progress and support of DRS in WES will be added in a future release of WES.
   - name: serviceinfo_model
     x-displayName: ServiceInfo
     description: |
@@ -150,7 +150,7 @@ paths:
       tags:
         - Service Info
       summary: GetServiceInfo
-      description: May include information related (but not limited to) the workflow descriptor formats, versions supported, the WES API versions supported, and information about general service availability.
+      description: Includes information related, but not limited to, the workflow descriptor formats, workflow engines, versions supported, the WES API versions supported, and information about general service availability and access.
       operationId: GetServiceInfo
       parameters: [ ]
       responses:
@@ -251,25 +251,25 @@ paths:
 
         **Content Types Supported:**
 
-        1. **`application/json`** (Recommended for single workflows submitted by URL): Submit a complete `RunRequest` object with structured data and type validation.  The advantage of this approach is the schema is explicitly defined by OpenAPI syntax (vs. just strings in multipart-form) along with workflow_unified_params object support (see below for more info).  Disadvantage is that it does not support workflow file uploads directly, which may be required for multi-file workflows.
+        1. **`application/json`** (Recommended for single workflows submitted by URL): Submit a complete `RunRequest` object with structured data and type validation.  The advantage of this approach is the schema is explicitly defined by OpenAPI syntax (vs. just strings in multipart-form) along with workflow_unified_params object support (see below for more info).  Disadvantage is that it does not support workflow file uploads directly, which may be required for multi-file workflows not available as URLs.
 
         2. **`multipart/form-data`**: Submit for cases requiring workflow file uploads (rather than just a URL). Fields are JSON-encoded strings that mirror the `RunRequest` structure.  Advantage is you can upload one or more workflow files directly while the disadvantage is you need to be careful in formating the JSON-ended string arguments correctly (multipart form submissions only support strings hence the limitation). 
 
         **File Handling:**
 
-        - **JSON requests**: Workflow file is referenced by URL or path. Use `workflow_unified_params` with File objects for rich metadata and generic parameterization.
+        - **`application/json` requests**: Workflow file is referenced by URL or path. Use `workflow_unified_params` with File objects for rich metadata and generic parameterization.
 
-        - **Multipart requests**: The `workflow_attachment` array uploads files required for workflow execution. Files are staged to a temporary directory with Content-Disposition headers containing filenames. Subdirectories are allowed but parent directory references ('..') are prohibited for security purposes.
+        - **`multipart/form-data` requests**: The `workflow_attachment` array uploads files required for workflow execution. Files are staged to a temporary directory with Content-Disposition headers containing filenames. Subdirectories are allowed but parent directory references ('..') are prohibited for security purposes.
 
         **Parameter Formats:**
 
-        - **`workflow_params`**: Language-specific parameters (CWL, WDL, Nextflow, SnakeMake format).  This field is used to send the native parameterization file (encoded as a string) used by the workflow engine.
+        - **`workflow_params`**: Language-specific parameters (CWL, WDL, Nextflow, Snakemake format).  This field is used to send the native parameterization file (encoded as a string) used by the workflow engine.
 
         - **`workflow_unified_params`**: Universal parameter format that WES server implementation converts to the target workflow language.  This allows for a more generic way to pass parameters that can be converted to the appropriate native format for the specified workflow_type.  This is an optional field, but if provided, it takes precedence over `workflow_params`.
 
         **Required Fields:**
 
-        - `workflow_type`: Workflow language ("CWL", "WDL", "Nextflow", "SnakeMake", or another alternative supported by this WES instance, see service-info)
+        - `workflow_type`: Workflow language ("CWL", "WDL", "Nextflow", "Snakemake", or another alternative supported by this WES instance, see service-info)
 
         - `workflow_type_version`: The workflow descriptor type version, must be one supported by this WES instance
 
@@ -296,7 +296,7 @@ paths:
                   description: JSON-encoded unified workflow parameters (see RunRequest for how to encode)
                 workflow_type:
                   type: string
-                  description: Workflow descriptor type (CWL, WDL, Nextflow, SnakeMake, etc. see service-info for supported types)
+                  description: Workflow descriptor type must be "CWL", "WDL", "Nextflow", or "Snakemake" currently (or another alternative supported by this WES instance, see service-info)
                 workflow_type_version:
                   type: string
                   description: The workflow descriptor type version, must be one supported by this WES instance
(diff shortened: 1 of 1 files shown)


uniqueg:
Thanks a lot for this effort, @briandoconnor!

But I need to be honest here: I always had a strong dislike for the `POST` Passport approach in DRS!

Apart from the security concerns which you already mentioned (sooner or later, someone _will_ end up logging the request - or, in this case, also the response - bodies), I think reserving `POST /objects` for this use case was a bad idea (and embarrassed our perfectly fine, useful and semantically sound custom `POST /objects` endpoint). Please don't let it be a precedent!

I guess there are cases where breaking REST semantics (and implications following from that, e.g., with respect to caching and idempotency) is an acceptable tradeoff. But in the case of passing around Passports, it just seems like a bad design decision on top of a bad design decision to me:

Ever since the April Connect in Switzerland, it was abundantly discussed and, IIRC, predominantly agreed with, that a generic mechanism by which assertions/visas could be provided back to a clearing house upon presentation of a _small_ Passport would constitute a cleaner design for Passports themselves, as well as for any APIs consuming them. Have sentiments changed since? If so, do you know what's the basis for the change of heart?

Besides, I think if credentials need to be passed around in the cloud, the state of the art seems to be the use of secret stores. In particular, HashiCorp Vault's API is open sourced and is a _de factor_ industry standard. Wouldn't it be better to explore an approach like that, which could be reused for all such cases (data access, compute access, tool access credentials in DRS, TRS, WES, TES etc.), rather than complicating schemas and duplicating endpoints across a whole bunch of API products individually?

vinjana:
Sorry for interfering, but I don't think it is a wise decision to adopt a "de facto industry standard" that is owned by a company for an open standard like the GA4GH standards are. There are numerous cases, where such companies arbitrarily changed their license (Akka, Anaconda, and you can probably find more) and suddenly users had to adapt. A standard should be grounded on a stronger foundation than the good will of a company. Sure, in both mentioned examples there were workarounds (Akka->Pekko, Anaconda->Mamba & Co.), but that does not mean they were trivial (the Anaconda licensing issue forced us to do lots of tests to ensure our workflows were backwards compatible). While such a change may be trivial for a standard -- just write "use the new fork of the old tool" -- the situation may be very different for the implementers and the operating institutions.

The Hashicorp Vault license says

> You may make production use of the Licensed Work, provided Your use does not include offering the Licensed Work to third parties on a hosted or embedded basis in order to compete with HashiCorp's paid version(s) of the Licensed Work.

This sounds vague and arbitrary. If Hashicorp decides to make their money in a different way that overlaps with the use of their tool in GA4GH we have to change the standard. This is the current license. So, I interpret that as that even for the current vault version a normal fork, like with Akka->Pekko, would not even be a legally solid option. I'm not a lawyer, sure, so I could be completely wrong.

uniqueg:
> Sorry for interfering, but I don't think it is a wise decision to adopt a "de facto industry standard" that is owned by a company for an open standard like the GA4GH standards are. There are numerous cases, where such companies arbitrarily changed their license (Akka, Anaconda, and you can probably find more) and suddenly users had to adapt. A standard should be grounded on a stronger foundation than the good will of a company. Sure, in both mentioned examples there were workarounds (Akka->Pekko, Anaconda->Mamba & Co.), but that does not mean they were trivial (the Anaconda licensing issue forced us to do lots of tests to ensure our workflows were backwards compatible). While such a change may be trivial for a standard -- just write "use the new fork of the old tool" -- the situation may be very different for the implementers and the operating institutions.
> 
> The Hashicorp Vault license says
> 
> > You may make production use of the Licensed Work, provided Your use does not include offering the Licensed Work to third parties on a hosted or embedded basis in order to compete with HashiCorp's paid version(s) of the Licensed Work.
> 
> This sounds vague and arbitrary. If Hashicorp decides to make their money in a different way that overlaps with the use of their tool in GA4GH we have to change the standard. This is the current license. So, I interpret that as that even for the current vault version a normal fork, like with Akka->Pekko, would not even be a legally solid option. I'm not a lawyer, sure, so I could be completely wrong.

You are not interfering, @vinjana - every voice is important!

I am not generally disagreeing with your important point, but I would like to clarify two points:
- I wasn't necessarily saying that the GA4GH should recommend a particular secret store API for sharing credentials, but rather that it's worth checking how similar use cases are handled by the wider community, especially given that passing secrets around in (multi-)cloud environments is not a problem that is particularly close to GA4GH's core focus (and expertise). What the outcome of such research and discussions will be, who knows? Maybe a document defining a generic process (e.g., "use a secret store") with examples, the recommendation of a specific API (such as HashiCorp's Vault API), a new API specification, or something entirely different (like this PR)...
- Unlike the HashiCorp's Vault _implementation_, the actual _API_ is (still) released under a permissive license approved by the OSI, [MSL-2.0](https://opensource.org/license/mpl-2-0), to be precise (see [here](https://github.com/hashicorp/vault/blob/main/api/LICENSE)). The same goes for much of the tooling that has been built around Vault, both by HashiCorp and others.

I would also like to note that there were previous discussions within the GA4GH community around potentially recommending existing standards in favor of drafting new ones, in order to avoid [this situation](https://xkcd.com/927/). Personally, I think that makes a lot of sense, especially in areas that are not intrinsically related to life sciences, genomics, healthcare etc. AFAIK, no official decision has been made or policy drafted to define rules around recommending external standards, e.g., with respect to the licensing terms, governance, revising recommendations etc. In this regard, the Vault API may be as good an example as we might get (at least in the Cloud WS) as a driver for such a discussion. FWIW, I wholeheartedly agree with you that we should probably _not_ recommend an API spec or other standard that is _not_ published under an OSI-approved license.

openapi/workflow_execution_service.openapi.yaml
```diff
@@ -820,7 +1291,27 @@ components:
...
+          description: >-
+            The outputs from the workflow run (legacy format). 
+            For outputs that require authentication to access (such as restricted files or directories), 
+            use the 'output_credentials' section of this response to obtain the necessary authentication 
+            credentials. The mapping between specific output files and credentials is implementation-specific 
+            for this legacy format - use 'structured_outputs' for explicit credential references.
+        structured_outputs:
+          $ref: '#/components/schemas/WorkflowOutputs'
+          description: Structured workflow outputs with rich metadata and type safety
+        output_credentials:
```
patmagee, line 1303: I do not think we should be embedding output credentials in the outputs, ideally we should NEVER have to return credentials but at most use something like a signed URL.

There are a few concerns here
1. Much less secure and exposes access credentials. I think returning DRS ids and then having the caller call each individually is moderately better, but even that pattern is not ideal
2. The engine may not be able to generate output credentials
3. Output credentials can be difficult to produce or lock down scope. IE a passport, bearertoken, api key could allow the user to access much more then JUST the outputs of the current run. A lot of DRS apis returned signed URls but these would be expensive to produce for a large workflow

patmagee (commented):
I can understand the challenge that this PR is trying to solve, however I fall in line with @uniqueg opinion here. I think breaking REST semantics is essentially a bad idea that will make the API harder to implement but also much less intuitive for the end user. IN particular I think making endpoints polymorphic makes writing clients in certain languages harder and also makes machine Interoperability harder IMO. If we really want to support `POST` then we may want to propose a totally separate RPC based api where these type of semantics are more accepted.

Additionally, If Passports are simply not usable within the context of a RESTful API service, it feels like we should be rethinking passports (similar to what @uniqueg mentioned as well), instead of propagating a bad design paradigm. Internally we use [token-exchange](https://oauth.net/2/token-exchange/) extensively which dramatically reduces the size of tokens that we pass around. Additionally, I know clouds have essentially the same thing, they typically call them things like "Workload Identity Federation". These are not vender specific but largely are using token exchange under the hood

- [GCP](https://cloud.google.com/iam/docs/workload-identity-federation)
- [AWS](https://aws.amazon.com/identity/federation/)
- [Azure](https://learn.microsoft.com/en-us/entra/workload-id/workload-identity-federation)

These are not vendor specific solutions and I would highly urge us to evaluate standards that are being developed by the broader software industry as they become available. Passport was initially created before things like token-exchange were adopted, but maybe we should revisit (not to use necessarily, just for inspiration.


Fundamentally, this PR is trying to achieve granting the engine access to data from a federated source using a passport or some other access credential.  I think that is a hard problem but maybe for a different reason.

Unlike TES, which is low level and can actually define how a single task interacts with underlying data (Ie it can write the command to use the credential), WES is a high level abstraction that in almost all cases sits in front of an engine. Any functionality we specify WES needs to fall into two categories of support

1. Something the engines are willing to adopt
2. Something the WES Server can do itself before handing off to the WES engine.

Data access is a hard one. I have done it in a similar way that is proposed here in a WES service. Essentially you end up needing to copy the data to a bucket from within your server or have some other workflow / batch job run. While this is technically doable (i did it), it was slow, error prone and caused data duplication. I could not read the data in-place once in the workflow.

On the flip side, we could not actually do this at the engine level because none of the engines supported this type of resolution and we are not an engine author ourselves. Additionally, most engines (at least for WDL) do not have a way of handling secrets, which basically means passing in sensitive information to the workflow itself should be avoided.

I don't think this means we cannot do it. BUT I do think we often are overlooking the key element here... the engine. A lot of WES requires the feature set provided by the engines

patmagee:
The more I think about the problem of Passport credentials, the more I realize we are actually on a very troubling path that is taking us away from established security best practices and current trends in the software industry.

Putting aside the challenges of overloading endpoints and breaking restful semantics, using a POST body to send credentials is problematic in the general case

- The body of a post request is much more likely to be logged then the headers
- Easily exposed in error messages 
- Monitoring tools (ie APM) may capture the values in the body
- Passes through multiple layers unencrypted: Once TLS is terminated the body is plaintext and the credentials are usable by anyone who wants to see them. Headers are also plain-text, so this just means that the attack surface has ballooned
- This now becomes easy to store in the DB accidentally and send back in the `RunLog` response

Additionally, the industry is largely going passwordless or putting a strong emphasis on platform provided credentials. There is a broad recognition that passing around ANY credential is typically a weak leek in the system so the best way to mitigate that is to simply not provide the raw credential to the user. Many clouds will now provide Credentials to their VM's without you having to make them yourselves, allowing the user to assign permissions to the VM's identity instead of passing in a particular credential. Additionally, most clouds have rolled out Workload Identity Federation allowing you to do OAuth2 token-exchange dramatically reducing the attack surface by only passing around a single credential that can be exchange by trusted client for the credentials needed to access other resources. Introducing a new approach where we are running the opposite direction to this feels like it has the potential for some serious mixups

I also think there is a general problem in how we are representing passport in general. The credentials that are being passed in are meant for OTHER services and are not meant for the WES service. This introduces a variety of challenges:

- Replay: Hopefully those credentials are short lived (although due to the nature of workflows, they likely are not), but they can be used by the current engine, or passed around without any difficulty. It imposes the same problem that 
- No validation: The credentials are meant for other services but we are just passing them around and not validating that they were actually intended to be sent to the WES service. In the OAUTH world this would come from validating that 1) the token has the appropriate audience and 2) comes from a trusted issuer and 3) is cryptographically verifiable. There are other assertions (ie scope, or custom ones) that can be made
- No Trust: This is a decentralized system and that part I understand, however I do not think we should be advocating for obliterating trust entirely. I think the server (WES) should have SOME sort of relationship with each target OAUTH server (ideally as an OAuth client). This would open the door for the Passport tokens to be VERY short lived, targeted tokens that are specifically meant for the WES client which the WES client can then exchange with the targeted issuer for a longer lived credential. This approach would actually eliminate the problem of eavesdroppers stealing the credential if tro exchange it the WES client needed to use it's client secret.

I think we need a broader discussion including security experts about how we should implement passport. My Gut feeling as that the current trajectory is taking is farther and farther away from established best practices and I would strongly advocate for deviating too far from them.

Issue #217, opened 2024-04-29
Consider using externalized common logging base model for WES and TES


uniqueg:
## Problem

Using the same schemas or base schemas across APIs to support identical or similar use cases facilitates implementation and therefore has the potential to increase adoption.

However, even though the [WES](https://github.com/ga4gh/workflow-execution-service-schemas) [`Log`](https://github.com/ga4gh/workflow-execution-service-schemas/blob/226f43b1fec97ad3778897a27a9651ccfa631d77/openapi/workflow_execution_service.openapi.yaml#L824-L866) schema and the [TES](https://github.com/ga4gh/task-execution-schemas) [`tesExecutorLog`](https://github.com/ga4gh/task-execution-schemas/blob/117cd927aa119469ee80cedbd8940c968b81c5d6/openapi/task_execution_service.openapi.yaml#L390-L431) schema are very similar to one another and likely originate from the same ancestor, they diverged over time. While this may well be for good reasons, it is also plausible that the divergence is simply a result of largely different communities working separately on the further development of the different APIs.

It might thus be worthwhile to explore whether the WES `Log` and TES `tesExecutorLog` schemas could be harmonized and how that could simplify or otherwise benefit the specifications.

## Possible solution

A possible solution is to replace WES `Log` and TES `tesExecutorLog` with a new schema that is defined in an independent, external OpenAPI document that could be maintained by the Cloud API and/or DaMaSC communities. An alternative to using the same identical schema for WES `Log` and TES `tesExecutorLog` could be to instead define a base schema that both schemas inherit from and extend differently.

The differences between the specifications are highlighted in the "Additional context" section below.

## Possible alternatives

If differences and/or use cases between both schemas are too different, it would not make sense to harmonize schemas. In that case, nothing should be done.

#### [WES](https://github.com/ga4gh/workflow-execution-service-schemas) [`Log`](https://github.com/ga4gh/workflow-execution-service-schemas/blob/226f43b1fec97ad3778897a27a9651ccfa631d77/openapi/workflow_execution_service.openapi.yaml#L824-L866) schema

```yaml
Log:
  title: Log
  type: object
  properties:
    name:
      type: string
      description: The task or workflow name
    cmd:
      type: array
      items:
        type: string
      description: The command line that was executed
    start_time:
      type: string
      description: When the command started executing, in ISO 8601 format "%Y-%m-%dT%H:%M:%SZ"
    end_time:
      type: string
[...]

PR #238, opened 2026-03-24
Extract common log base schema for WES/TES harmonization

knQzx:
WES `Log` and TES `tesExecutorLog` share the same core fields (start_time, end_time, stdout, stderr, exit_code) but diverged over time into separate definitions. this makes it harder to build clients that work with both APIs extracted the shared fields into `common_log_base.yaml` as a `LogBase` schema, and refactored WES `Log` to extend it via allOf. TES can reference the same base on their side

the descriptions are unified to cover both use cases — WES uses URLs for stdout/stderr while TES may return content directly, so the base description accounts for both

relates to #217

I emailed gsoc-mentors-2025@ga4gh.org about GSoC 2026 (knqzx0@gmail.com)

### DIFF
--- a/openapi/common_log_base.yaml
+++ b/openapi/common_log_base.yaml
@@ -1,0 +1,37 @@
+components:
+  schemas:
+    LogBase:
+      title: LogBase
+      description: >-
+        Base log schema shared between WES Log and TES tesExecutorLog.
+        See https://github.com/ga4gh/workflow-execution-service-schemas/issues/217
+      type: object
+      properties:
+        start_time:
+          type: string
+          description: >-
+            When the command started executing,
+            in ISO 8601 / RFC 3339 format
+          example: "2020-10-02T10:00:00-05:00"
+        end_time:
+          type: string
+          description: >-
+            When the command stopped executing (completed, failed, or cancelled),
+            in ISO 8601 / RFC 3339 format
+          example: "2020-10-02T11:00:00-05:00"
+        stdout:
+          type: string
+          description: >-
+            Stdout content or a URL to retrieve it.
+            No guarantees are made about completeness - implementations
+            may return only head/tail or a URL reference.
+        stderr:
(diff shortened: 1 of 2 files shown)


patmagee:
hey @knQzx thank your for your contribution! It looks like you're targeting the master branch, but we should actually be targeting the develop branch. Would you mind changing the PR target? 

Overall, I like the idea of where you're coming from. But I'm not sure the proposal really solves the problem that you are thinking of. If the underlying issue is spec drift, extracting into a common library definitely helps, but that common library needs to be upstream of the West and test services and likely all referring to the exact same definition. 

Extracting it into a separate common schema without referencing the exact same schema between both api definitions doesn't necessarily help with client generation, maybe in untyped languages, but in typed languages it would still fail. 

Would love to hear your thoughts on that.

knQzx:
> hey @knQzx thank your for your contribution! It looks like you're targeting the master branch, but we should actually be targeting the develop branch. Would you mind changing the PR target?
> 
> Overall, I like the idea of where you're coming from. But I'm not sure the proposal really solves the problem that you are thinking of. If the underlying issue is spec drift, extracting into a common library definitely helps, but that common library needs to be upstream of the West and test services and likely all referring to the exact same definition.
> 
> Extracting it into a separate common schema without referencing the exact same schema between both api definitions doesn't necessarily help with client generation, maybe in untyped languages, but in typed languages it would still fail.
> 
> Would love to hear your thoughts on that.

thanks!! retargeting to develop now

yeah that makes sense, having it only in the WES repo doesn't help with typed codegen since TES would still have its own definition. i noticed WES already does cross-repo refs for service - info via raw github URLs, so probably the same approach would work here - put the base schema somewhere shared, version it, and both specs $ref to the same thing

this was mostly to get feedback on which fields make sense to share. happy to move it or close and reopen wherever fits

